PR #27, opened 2024-07-02
Fix symlinks

nakohdo:
The symlinks for the default Antora folders "pages" and "images" were broken so the actual content was duplicated. I removed the  folders and recreated the symlinks.


ClemensLinnhoff:
I would like to merge everything concerning the setup of the documentation page as soon as possible. These types of PRs are very hard to review anyways. @AsamDiegoSanchez can we expedite the merge of this without format CCB approval?

@nakohdo Will this get the GitHub pages docu page working? Or is there still something missing? I think there is already something there at https://asam-ev.github.io/OpenMATERIAL/.

nakohdo:
@ClemensLinnhoff 
No, this is just to fix the symlinks that somehow were broken again. This will only make the documentation conform to the required Antora folder structure as a prerequisite for the build process. 

I want to keep both things separate so there will be a separate PR for the publication on GitHub Pages. What you can see there was created from a test branch, not yet from "main".

ClemensLinnhoff:
Then @AsamDiegoSanchez let's merge this asap, so we can move on with the implementation.

ClemensLinnhoff (approved):

AsamDiegoSanchez:
Thank you @nakohdo.
Agree, let´s merge it so we can start.

PR #30, opened 2024-07-04
Add gitignore

FKlopfer:
## Describe your changes

Add a gitignore file, containing initial content that IDE specific files are ignored by Git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+# IDE specific files
+.vscode
+.idea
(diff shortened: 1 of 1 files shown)


ClemensLinnhoff (approved):
Thanks. I think this would be enough now. We can always add more later.

ClemensLinnhoff:
@AsamDiegoSanchez In this case I also think we can expedite the merge without formal CCB approval. What do you think?

AsamDiegoSanchez:
Agree, we can move forward

PR #56, opened 2024-08-05
Fix the table of content for geometry and material chapter

FKlopfer:
Update the antora nav file so that all subheadings are now in correct order


ClemensLinnhoff:
Minor editorial change: Merge without CCB approval.

PR #82, opened 2024-09-04
Add elasticity and density

### LINKED ISSUE #58: Add Density of materials to material properties
For modeling ultrasonic sensors you need to calculate the speed of sound in a material. This can either be added directly, but the group decided to add its components. One if which is the density of material.

Add density in kg / m^3 to the material properties.

### LINKED ISSUE #60: Define elasticity in the material properties
The simulation of ultrasonic sensors requires the speed of sound in and the acoustic impedance of materials.
These quantities can be calculated from density and elastic properties which are more generic properties that might have additional use cases.

For homogeneous isotropic materials any two of the following properties will suffice:
 - Young's modulus (E)
 - Poisson's ratio (&nu;) 
 - bulk modulus (K)
 - shear modulus (G) 

( see https://en.wikipedia.org/wiki/Young%27s_modulus#Usage **E = 2G(1+&nu;) = 3K(1-2&nu;)**)

Since bulk and shear modulus are easily derived, I would propose to **add Young's modulus and Poisson's ratio** to the material properties.

ipg-jsc:
## Describe your changes
As discussed in the meeting on 04.09.2024:
- [x] Add elasticity properties in form of Young's modulus and Poisson's ratio to material example
- [x] Add density

Closes #58, closes #60


ClemensLinnhoff (approved):
Looks good to me, thanks.
The only thing I am not sure about is which fields should be required. But I propose to go over all fields in the end and decide if they should be required or not.

ClemensLinnhoff:
@AsamDiegoSanchez @drsftx73 @KimuraDIVP could you do an additional formal review so we have the full support of all relevant CCB members?

KimuraDIVP (approved):

ipg-jsc:
@ClemensLinnhoff have you had anything specific in mind on how to add the [source field](#95) to properties that are currently a single key-value pair?

I could imagine going for @DavidJRitter904's approach in #86 and call the higher level key `density_data`.
For the sake of consistency I would also change `elasticity` &rarr; `elasticity_data`.

```json
    "elasticity_data": {
        "youngs_modulus": 70e9,
        "poissons_ratio": 0.35,
        "source": "internet: https://en.wikipedia.org/wiki/Aluminium"
    },
    "density_data": {
        "density": 2699.0,
        "source": "internet: https://en.wikipedia.org/wiki/Aluminium"
    },
```

ClemensLinnhoff:
> @ClemensLinnhoff have you had anything specific in mind on how to add the [source field](#95) to properties that are currently a single key-value pair?
> 
> I could imagine going for @DavidJRitter904's approach in #86 and call the higher level key `density_data`. For the sake of consistency I would also change `elasticity` → `elasticity_data`.
> 
> ```json
>     "elasticity_data": {
>         "youngs_modulus": 70e9,
>         "poissons_ratio": 0.35,
>         "source": "internet: https://en.wikipedia.org/wiki/Aluminium"
>     },
>     "density_data": {
>         "density": 2699.0,
>         "source": "internet: https://en.wikipedia.org/wiki/Aluminium"
>     },
> ```

I think this is a very good solution! I originally thought about something like this:

```json
"density": {
    "value": 2699.0,
    "source": "internet: https://en.wikipedia.org/wiki/Aluminium"
},
```
But I actually like your approach better.

PR #97, opened 2024-09-12
update validate-json.yaml

mettelilienthal:
This should improve the json validation display by showing where the error is by line and column.

### DIFF
--- a/.github/workflows/validate-json.yaml
+++ b/.github/workflows/validate-json.yaml
@@ -1,6 +1,6 @@
 # Validate JSON schema files
-# Source: https://github.com/GrantBirki/json-yaml-validate
-# Forked from: https://github.com/marketplace/actions/json-yaml-validate
+# Source: https://github.com/dsanders11/json-schema-validate-action
+# Forked from: https://github.com/marketplace/actions/json-schema-validate
 name: Validate JSON schemas for assets and materials
 
 on:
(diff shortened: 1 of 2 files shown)


ClemensLinnhoff:
Why is it checking both for YML and JSON when we only have a json input?

image

Also, the JSON part actually does not give a position, just the YML part but the line is actually not completely correct.

ClemensLinnhoff:
And please remember to add your sign-off so the DCO does not fail. It is just a `-s` on your commit.

mettelilienthal:
@ClemensLinnhoff we've removed the yaml validation.

ClemensLinnhoff:
> @ClemensLinnhoff we've removed the yaml validation.

Thanks, this is better, but now it does not show in which file the error occurred, as far as I see.

image

martinfiebig:
Hi @ClemensLinnhoff,
we have now tested all available GitHub Actions on GitHub Marketplace for the validation of JSON files. In our tests, none of the available Actions could satisfy all the needs (display line/column, display the file causing the error, validate against JSON only).
We concentrated on GitHub Actions because adding additional custom code (i.e. JavaScript or python files) to the ASAM repository would lead to higher maintenance effort.

The following GitHub Actions performed best while testing - althoug they have different disadvantages:

1. [GitHub Action by 'dsanders11'](https://github.com/marketplace/actions/json-schema-validate) (no display of the file causing the error):
   image

2. [GitHub Action by 'cardinalby'](https://github.com/marketplace/actions/schema-validation-action) (needs to run both JSON and YAML validation):
image
For the JSON part, the Action still only gives a position, because the JSON input is parsed as a one-line-item. For the YAML part, the input is parsed differently so the action can display the line and column of the error.
   
Please let us know which one of the Actions above is acceptable and should be implemented, or if we should switch back to the original solution.

ClemensLinnhoff:
Okay thank you for testing. Then let's go with the second option. It has the name and the position specified (the yml part actually adds the line info.).

martinfiebig:
Hi Clemens, thank you for the quick reply. I've implemented the second GitHub Action now.

ClemensLinnhoff (approved):

PR #101, opened 2024-09-12
add retro-reflectivity to the material schema

### LINKED ISSUE #5: Add retroreflective material
Original author: @ClemensLinnhoff 

Especially for lidar, retroreflective materials such as road signs and license plates are important.

I suggest either to include a flag if a material is retroreflective for a certain wavelength(region).
Or to be more precise, we need to introduce a "retroreflectivity" value, because no material is perfectly retroreflective. If it was, road signs would be dark for the human eye. So there is always a diffuse portion.
The problem is, I don't have a good idea how to physically set this parameter.

ipg-jsc:
## Describe your changes

Add coefficient of retro-reflection as optional material property as discussed in the meeting on 11.09.24 and in issue https://github.com/asam-ev/OpenMATERIAL/issues/5.

Closes #5 


ClemensLinnhoff (approved):

KimuraDIVP:
Are specular-reflectivity and diffuse-reflectivity defined somewhere alse?
I'm not quite sure why only retro-reflectivity commented here.

ClemensLinnhoff:
The clear differentiation between diffuse and specular reflectivity is not that easy, as there are also mix form.
Retroreflectivity is a special case, as it is defined by regulators e.g. for traffic signs and license plates.

KimuraDIVP (commented):
o.k.

PR #112, opened 2024-09-19
Add example vehicle

### LINKED ISSUE #111: Add example vehicle to demonstrate vehicle node structure
Having a concrete example to showcase the standard definition is always helpful.

Modify the existing example vehicle from the legacy OpenMATERIAL to reflect the current standard description.

ClemensLinnhoff:
## Describe your changes

Add example vehicle in glTF format to showcase the definitions of the vehicle node structure.

- [x] Fix structure according to new definitions
- [x] Fix huge bevels at doors (all edges) and the convertible top (front/tail edge)

Fixes #111 


ClemensLinnhoff:
I updated the 3D model to the current definition of the vehicle structure. Please review the internal structure, node naming, coordinate origins (pivot points) etc.
I also added a corresponding asset file (without material mapping).

AsamDiegoSanchez (approved):
General check, looks ok to me.
Signed-off-by: Diego Sánchez <diego.sanchez@asam.net>

LudwigFriedmann:
First remarks (WIP)

**Functional:**
* [Problem in the structure]
Side mirrors and mountings are not affected by door animation.
They could be either mounted to the fenders or they have to become children to the doors.
In my opinion, both cases are relevant and should be considered in the structure.
For this example, I'd prefer to have the mirrors mounted to the doors.
* [Problem in the structure]
Coordinate frames of the side mirrors are positioned in the mirror mounting joints, not in the center of the mirror surface.
This is helpful for animation of mirror adjustment, but a coordinate for reflections or mirror cameras is missing.
A solution could be an additional child coordinate frame in the center of the mirror surface, aligned with the tangens/normal of the mirror.
* ...

**Artistic;**
* Bevels of doors (all edges) and the convertible top (front edge) could be reduced
* ...

ClemensLinnhoff:
Thanks @LudwigFriedmann, let's discuss your remarks in the meeting tomorrow (especially the conceptual issues).

ClemensLinnhoff:
> * [Example issue]
>   Lateral iterators are inconsistent for low beam lights (0 = left) and wheels (0=right).
>   This may be true for other nodes, as well.

I checked this again and I think it is correct in the example. We only have one Light_Low_Beam_Left. So the index 0 is correct in my opinion.
The index for the wheels also is correct, as we count in positive y-direction from right to left.

Do you agree @LudwigFriedmann ?

LudwigFriedmann:
@ClemensLinnhoff : You're correct, I missed the "Left" in Light_Low_Beam_Left and misinterpreted the 0 as a lateral iterator.
In constrast, wheels feature a lateral iterator and no "left" or "right".
Should we harmonize that?

ClemensLinnhoff:
> @ClemensLinnhoff : You're correct, I missed the "Left" in Light_Low_Beam_Left and misinterpreted the 0 as a lateral iterator. In contrast, wheels feature a lateral iterator and no "left" or "right". Should we harmonize that?

The wheels are harmonized with OSI, so I would not change their definition.
For the lights I think we defined them as left and right  because they are typically on either side of the vehicle. So in my opinion it would be fine to stick with the current definition.

ClemensLinnhoff:
I updated the example vehicle according to the review remarks and the changes to the node structure introduced in #171.
@LudwigFriedmann please re-review.

LudwigFriedmann:
Hi @ClemensLinnhoff, thanks for all the improvements.

I have a few minor remarks, only:

[Concept issue]
We should have a note somewhere that only node names are subject to standardization.
For meshes, arbitrary names can be chosen.

[Concept/Example issue]
For doors and lights, we use suffixes "Left" and "Right".
For Wheels, we use longitudinal and lateral iterators.
Seats introduce a new, 2-dimensional iterator concept.
At least the seat iterarators should be harmonized.

[Artistic issue]
There are still some rendering problems at
* front and back edges of the convertible top
* inner edges of the trunk lid
* edges around the rear lights

This happens because "Shade smooth" is enabled for the smooth parts.
As Blender doesn't know which edges are meant to stay sharp, it tries to all edges by introducing bevels.
For sharp edges and contact surfaces/edges, specifying "Sharp edges" should solve this problem.

ClemensLinnhoff:
We fixed the artistic issue. But we had to do a complete remesh, because there were too many imperfections from the original version. Now it should be fine.

ClemensLinnhoff:
I added the rear axle center to the definition as well as the glTF.

Now just the seat index is open. @LudwigFriedmann I don't remember if we made a final decision about this in the meeting.

LudwigFriedmann:
Hi @ClemensLinnhoff, as far as I remember, the most favored solution was to have a row index and a row-specific seat index.

ClemensLinnhoff:
> Hi @ClemensLinnhoff, as far as I remember, the most favored solution was to have a row index and a row-specific seat index.

Okay thanks. I fixed the naming both in the definitions and in the glTF.

LudwigFriedmann (approved):
The car looks great and the documentation reflects this. Thanks for all the work!

PR #146, opened 2024-10-22
Add images for vehicle structure 

FKlopfer:
## Describe your changes

Adding images to illustrate vehicle structure

[PR 103](https://github.com/asam-ev/OpenMATERIAL/pull/103)


ClemensLinnhoff:
Here are my remarks from #103:

Please make sure, that the coordinate frame fit to the description of the specific element. For example the coordinate frome on the front door (engine hood) does not comply to the description.

Please also make sure that all illustrations are consistent.
- The Light_Day_Right looks completely different for the one on the left.
- Also it would be nice, if the coordinate frame had the same size in all figures.
- Grp_Door_Top does not fit to the style of the other images at all.

Could you change the color/grayscale of the coordinates so that they are visible both in light and dark mode?

content/_images/Grp_Mirror_Rearview_Mounting.svg
ACMackenthun, line 1: @ClemensLinnhoff: I updated the image and fixed the coordinate system. If x points forward and the z-axis is perpendicular to the x-axis (or rather to the x-y-plane to make it less ambiguous), then the y-axis in the right-handed coordinate system would point to the left.

Can you please doublecheck and confirm whether I interpreted the definition correctly? We will then update the other interior images.

ClemensLinnhoff: Yes, that looks correct.

content/_images/Grp_Vehicle_Part.svg
ACMackenthun, line 1: I added this image of a hinged bus to explain the orientation of vehicle parts. However, it was not easy to visualize that the y-axis points to the left in driving direction of the part in the back. We might also need to update the image as soon as the working group has decided on the origin of the coordinate system for vehicle parts.

@ClemensLinnhoff: Can you please check whether you agree with the visualization of the coordinate system?

ClemensLinnhoff: The bus itself looks very nice (although I think the front part needs to have two axes).
For the coordinate system, as you already mentioned, the y-axis looks a bit weird. Does not seem to be in a right angle to the other axes.
But we need to decide first, where exactly the origins of the vehicle parts should be, see #148.

ACMackenthun: Thanks for the fast reply. You are right about the second axle. I missed that it he first shot, but added it now.

We will update the coordinate system as soon as there is a decision on its origin. Maybe we could improve it by changing the arrow direction of the y-axis to lower left as below, where I rotated the y-axis by 45° to the left of the x-axis.

image

content/_images/Grp_Mirror_Blindspot_Mounting.svg
ACMackenthun, line 1: I reused the bus image to illustrate the blindspot image mounting. However, the 3D perspective makes it hard to visualize the axes of the coordinate system. In this case, the z- and y-axes are easy, but I had issues with the rotation of the x-axis.

@ClemensLinnhoff: Can you please check the rotation of the x-axis and the origin of the coordinate system (I interpreted "base" as lowest point of the mounting, but it could also refer to the centre)? We could also create a new image from the bus driver's perspective, which could be reused in Grp_Mirror_Blindspot_View.

ClemensLinnhoff: That looks good.

content/_images/Grp_Mirror_Side_View_Left.svg
ACMackenthun, line 1: @ClemensLinnhoff: Please doublecheck. I fixed the x-axis, which pointed to the front in the previous version, and also had to change the y-axis to the inside of the car to create a correct right-handed coordinate system.

We originally misinterpreted the definition "Pointing outwards from the front of the mirror". I propose to rephrase the complete definition of the x-axis, e.g. to "Pointing back following the reflection from the mirror surface". Or do we have to be even more specific and include the angle of the mirror surface or the relation to the viewer?

ClemensLinnhoff: It is not really pointing back following the reflection as the reflection would depend on the position of the observer. More accurate would be that the x-axis points in the direction of the face normal of the mirror surface.

content/_images/Grp_Door_Front.svg
ACMackenthun, line 1: @ClemensLinnhoff: The definitions of front door, rear door, and top door mention that the z-axis is "concentric and coaxial to the virtual hinge axis", but they did not mention whether they point to the left or right, which made it hard to identify the direction of the y-axis. Can you please doublecheck the directions of the axes?

content/_images/Grp_Mirror_Side_View_Left.svg
ACMackenthun: Thanks for the fast reply. I will update the definition.

content/_images/Grp_Door_Front.svg
ClemensLinnhoff: You are right. I would propose to orient the z-axis, so that the door opens in a right handed coordinate system. So having a positive angle around the z-axis would open the door. This means on a left door it would point down and on a right door it would point up.

Could you add this to the definitions?

Btw. I just notices, that the x-axis on the Grp_Door_Left is not correct. It should point inside the door. The other doors are correct.

ACMackenthun: Yes, I will update the definitions. Thanks for the clarificiation.

I also just committed an updated version of the Grp_Door_Left image with updated x- and y-axis.

ClemensLinnhoff: The x-axis is correct now. But with the new definition, the z-axis should point down.

ClemensLinnhoff:
The chapter structure was altered by #159 as the nodes were sorted alphabetically. This will potentially lead to merge conflicts here. So please merge the main branch into this one and fix all potential conflicts.

FKlopfer:
Main branch is merged into this branch and merge conflicts solved.

content/_images/Grp_Door_Bottom.svg
ACMackenthun, line 1: I added this image, but it is quite hard to imagine a vehicle with a door at the bottom. I did some research and expect that this is intended for vehicles with separated tailgates, e.g. a pickup truck with a hatch that folds down to the back. If that is the case, I am wondering about the x-axis, which points to the front, because we defined that doors rotate around the z-axis, thus moving between x- and y-axis.

Is my interpretation of the door location correct or do you really refer to a hatch at the bottom of a vehicle?

ClemensLinnhoff: You are right, a standard road vehicle probably would not have such a door. We just added it for a complete definition and for potential off-road vehicles, that might have something like an escape hatch at the bottom.
So for the image I would prefer to have a perspective, where you can actually see the bottom of a car rather than the rear.

ClemensLinnhoff:
> Main branch is merged into this branch and merge conflicts solved.

Just as a heads up: PR #170 might introduce more merge conflicts. But nothing big. We will just add a new hierarchy level for the mirrors and added some more detail to some of the descriptions. But the general layout stays the same.
There we also fix the definition of the vehicle parts. We now are making them optional to be defined in future releases. But keeping the image of the articulated bis is still a good idea to visualize what a vehicle part could be.

content/_images/Grp_Door_Bottom.svg
ACMackenthun: I added a simplified image with a different perspective. Is the hinge axis realistic or do you prefer a different edge?

content/_images/Grp_Door_Front.svg
ACMackenthun: I updated the Grp_Door_left image and the z-axis points down in the new version

content/_images/Grp_Door_Bottom.svg
ClemensLinnhoff: That looks good, thanks.

ACMackenthun (approved):
@ClemensLinnhoff: I doublechecked the images and the image references in the *adoc file. 

All of the requested images are available and added to the spec and I opened the PR for the official review. Could you please have a final look?

Please note that I did not yet update the definition of the z-axis for doors, because this PR focused on the images. If you don't mind, I will use https://github.com/asam-ev/OpenMATERIAL/pull/171 for updating the definition.

content/_images/Grp_Door_Top.svg
ClemensLinnhoff, line 1: Similar to the door on the bottom, could you make an image, where the door is more on the top?

content/_images/Grp_Light_License_Plate.svg
ClemensLinnhoff, line 1: The x-axis looks a bit crooked here.

content/_images/Grp_Light_Warning.svg
ClemensLinnhoff, line 1: The x-axis looks a bit crooked here, as well.

content/_images/Grp_Mirror_Blindspot_View.svg
ClemensLinnhoff, line 1: The view direction should be the surface normal of the mirror glass, so pointing backwards here.

content/_images/Grp_Mirror_Rearview_View.svg
ClemensLinnhoff, line 1: The view direction should be the surface normal of the mirror glass, so towards the the viewer with the y-axis to the right.

content/_images/Grp_Vehicle_Part.svg
ClemensLinnhoff, line 1: In the new definition the origin of the vehicle part is in the connection points to the other vehicle part. So if the vehicle part in this example is the rear part of the articulated bus, the origin should be in the center of the hinge, where it is connected to the front part. Maybe you can make the front in a light gray to indicate, that here only the rear part is considered the vehicle part in question.

ClemensLinnhoff (changes requested):
Thanks, most of it looks very good! I only have minor remarks.
And I agree, let's do the changes to the definitions in #170. And also the new nodes for the mirror joints can be illustrated there.

content/_images/Grp_Door_Top.svg
ACMackenthun: I added a new car model with a roof door. We could also use this image for the convertible top class.

The perspective of this car would also be easier to visualize [Grp_Light_License_Plate.svg](https://github.com/asam-ev/OpenMATERIAL/pull/146/files/f30a9afd836781128fe961c2621f29112fe1ae42#diff-5eefdd34cb239e7d600d6cb15023587736f57c5f63776f11400f6de6cbb0b8ab) and [Grp_Light_Warning.svg](https://github.com/asam-ev/OpenMATERIAL/pull/146/files/f30a9afd836781128fe961c2621f29112fe1ae42#diff-0dae49ebede4284375276b9fe56064499a0c0375a963b46fcb6696ead059e5d1). I will provide updated images for these examples tomorrow

content/_images/Grp_Mirror_Rearview_View.svg
ACMackenthun: Updated image

content/_images/Grp_Vehicle_Part.svg
ACMackenthun: Updated image. I used light gray for the front part and put the coordinate system at the center of the virtual hinge axis. Please check.

content/_images/Grp_Light_License_Plate.svg
ACMackenthun: I will provide a new version based on the roof door image tomorrow

content/_images/Grp_Light_Warning.svg
ACMackenthun: I will provide a new version based on the roof door image tomorrow

content/_images/Grp_Light_License_Plate.svg
ACMackenthun: I updated the image with the new perspective, which makes it easier to visualize the x-axis

content/_images/Grp_Mirror_Blindspot_View.svg
ClemensLinnhoff: @ACMackenthun I think this is the last thing that needs to be fixed. After this we could merge.

ACMackenthun: Sorry, I missed that one. I fixed the x- and y-axis in the image. The perspective is not ideal, because it would be easier to visualize the angles from the point of view of the driver, but I guess that the reader understands the image in combination with the text.

ClemensLinnhoff: Thanks, then we are ready for review :)

ClemensLinnhoff (approved):

ClemensLinnhoff:
We now already merged #170. 
@FKlopfer could you merge the main branch back into this one and fix potential merge conflicts?
@ACMackenthun there is an additional "joint" node for each mirror (left, right, blindspot and rearview). Could you add illustrations for these nodes here, too? The vehicle parts node was moved all the way to the end of the definitions. Please check, that the illustrations there is still correct.

content/_images/Grp_Mirror_Blindspot_Joint.svg
ACMackenthun, line 1: @ClemensLinnhoff: I added the images for the mirror joints and slightly updated the mirror mounting images to visualize that the origin differs, because the mounting origin is located at the base and the joint origin at the joint itself.

Hope that the difference is clear enough for the target audience..

content/geometry/object-vehicle/vehicle-structure.adoc
```diff
@@ -1614,15 +1612,15 @@ NOTE: Add a figure.
 |===
 
 
-== Grp_Mirror_Rearview_Joint_Left_<rearview_mirror_joint_idx>
+== Grp_Mirror_Rearview_Joint_<rearview_mirror_joint_idx>
```
ACMackenthun, line 1616: I noticed a typo in the section title and removed the "_Left".

@LudwigFriedmann: Please doublecheck whether that change was correct.

ClemensLinnhoff: This is indeed a typo. I thought I fixed this already. Hm maybe not. Thanks!

content/_images/Grp_Mirror_Blindspot_Joint.svg
ClemensLinnhoff: Could you make this a bit clearer, at least on the side mirrors? Mostly the joint is further inwards:

image

@LudwigFriedmann what do you think?

ACMackenthun: No problem, I just committed a new version of both side mirror images where I moved the coordinate systems to the proposed position

ClemensLinnhoff: Looks good, thanks.

ClemensLinnhoff:
This PR is ready to be merged from my point of view.
I leave it to @LudwigFriedmann for the final review.

LudwigFriedmann (approved):
Looks good to me - thanks for all the work!

PR #158, opened 2024-10-25
157 provide example for a human asset analog to vehicle example

### LINKED ISSUE #157: Provide example for a human asset analog to vehicle example
A vehicle example is introduced in #112. Examples for humans and environments are still missing.

Provide a glTF file with a human model.

ClemensLinnhoff:
## Describe your changes

Add human model (without helper bones and with full hands (no fingers as bones))

Fixes #157 

## Mention a member

@ipg-sig 

- [ ] I have performed a self-review of my code/documentation.
- [ ] My changes generate no new warnings during the documentation generation.


ipg-sig:
@ClemensLinnhoff 

Should the included animation look like this or wasn´t it intended?
Anim

ClemensLinnhoff:
> @ClemensLinnhoff
> 
> Should the included animation look like this or wasn´t it intended?

I think we had some animations in there (running, jumping or similar). But they were not intended to be exported as part of this example. Feel free to remove them.

LudwigFriedmann:
@ClemensLinnhoff Is this PR ready for review?

ClemensLinnhoff:
No, this is still a draft. This was just my initial upload and @ipg-sig might need to adapt the model to the new definitions.

ClemensLinnhoff:
I think there is something wrong with the node hierarchy. I imported the glTF into Blender and the hierarchy looks like this:

image

Looks like the hierarchy was flattened. Was that done during export maybe?

ClemensLinnhoff:
Otherwise it looks great. I like the example of the wrist band as an accessory.

LudwigFriedmann:
It may be the case that blender and/or the gltf exporter considers nodes which are linked to the armature as children.
I can export my previous human model and see if I experience similar results

ClemensLinnhoff:
I checked the updated human example and the hierarchy is still flattened in Blender. But maybe this is a Blender bug?
Because this is an excerpt of the current glTF file and here the group nodes do have children.

```json
{
	"mesh":0,
	"name":"Accessorie",
	"skin":0
},
{
	"children":[
		31
	],
	"name":"Grp_Accessories"
},
{
	"mesh":1,
	"name":"Human",
	"skin":0
},
{
	"children":[
		33
	],
	"name":"Grp_Body"
},
{
	"mesh":2,
	"name":"Vest",
	"skin":0
},
{
	"children":[
		35
	],
	"name":"Grp_Clothing"
},

```

ClemensLinnhoff:
And I get the following report from a glTF validator:

image

ClemensLinnhoff:
On the other hand, I imported it to this editor: https://www.gltfeditor.com/

Here the structure looks correct:

image

ClemensLinnhoff:
I also asked our 3D artist and he says it should not be possible to have empty nodes with children in an armature. The meshes should directly be part of it, at least in Blender.

ClemensLinnhoff:
> I also asked our 3D artist and he says it should not be possible to have empty nodes with children in an armature. The meshes should directly be part of it, at least in Blender.

Okay we tried around a bit. So it does seem to work. But it is an import bug in Blender that flattens the hierarchy. Anyways this seems to be quite an exotic solution having empty nodes as part of an armature.

ipg-sig:
@ClemensLinnhoff Thank you very much! I have tried different combinations in Blender, but it always ended in the flattened hierarchy after reimporting the exported files (all 3 file formats).

To avoid it, we could also use the group names as an Prefix for the meshes, so we can get rid of the additional group nodes.
Example: 
- Accessories_Wristband
- Body_Human
- Clothing_Vest
- Hair_Short

But I think this should be discussed in the group next week?
I can fix it in advance for next week, but I will not be able to be present at the next meeting.

ClemensLinnhoff:
> @ClemensLinnhoff Thank you very much! I have tried different combinations in Blender, but it always ended in the flattened hierarchy after reimporting the exported files (all 3 file formats).
> 
> To avoid it, we could also use the group names as an Prefix for the meshes, so we can get rid of the additional group nodes. Example:
> 
> * Accessories_Wristband
> * Body_Human
> * Clothing_Vest
> * Hair_Short
> 
> But I think this should be discussed in the group next week? I can fix it in advance for next week, but I will not be able to be present at the next meeting.

Yes, this came also to my mind. Just have a prefix for the meshes instead of the empty nodes.

LudwigFriedmann:
I can confirm flattening of the hierachy happens for my example, as well (even though I've made sure this option is disabled at exports).

Blend:
pedestrian_blend

 glTF:
pedestrian_gltf

ClemensLinnhoff:
We created an issue in the Blender project and already got an answer from another user and from a blender organization member who do not think this is a bug since the behavior of the mesh is still correct even if the hierarchy is flattened. The issue was already closed:
https://projects.blender.org/blender/blender/issues/130677

So if Blender is not supporting this and will not support this, I think we should flatten the hierarchy in the definitions as well, as @ipg-sig proposed.

ipg-sig:
Model and exported files re updated @ClemensLinnhoff .

With OpenUSD the "Grp_Root" was always renamed to "root" after import, but I am not sure if this is another Blender OpenUSD Bug. FBX and GLTF work fine now. maybe someone can check it again with OpenUSD?
The .blend file is also updated.

examples/human_example/human_example.gltf
```diff
@@ -0,0 +1,2335 @@
...
+			"buffer":0,
+			"byteLength":25788,
+			"byteOffset":1603276,
+			"target":34963
+		}
+	],
+	"buffers":[
+		{
+			"byteLength":1629064,
+			"uri":"Human_Example.bin"
```
ClemensLinnhoff, line 2332: The file name is wrong here. This would be a simple fix, if you are sure you uploaded the correct files. Just want to make sure.

ClemensLinnhoff (commented):
Looks very good, just one minor remark.

ClemensLinnhoff:
@ipg-sig feel free to add yourself as co-author the the metadata of the asset file :)

LudwigFriedmann:
> Model and exported files re updated @ClemensLinnhoff .
> 
> With OpenUSD the "Grp_Root" was always renamed to "root" after import, but I am not sure if this is another Blender OpenUSD Bug. FBX and GLTF work fine now. maybe someone can check it again with OpenUSD? The .blend file is also updated.

I wasn't able to reproduce this behavior in Blender 4.1.1.
"Grp_Node" kept its name over an export and re-import as USD.
image

ClemensLinnhoff:
> > Model and exported files re updated @ClemensLinnhoff .
> > With OpenUSD the "Grp_Root" was always renamed to "root" after import, but I am not sure if this is another Blender OpenUSD Bug. FBX and GLTF work fine now. maybe someone can check it again with OpenUSD? The .blend file is also updated.
> 
> I wasn't able to reproduce this behavior in Blender 4.1.1. "Grp_Node" kept its name over an export and re-import as USD. 

I can somewhat reproduce it. After reimporting the pedestrian USD file, I get the following structure:

image

ipg-sig:
> Sometimes the x-axis is pointing forwards, sometime backwards. Can we unify this?

We could unify it, but we would lose the ability to mirror the animation properly. For example: If I have mirroring enabled and rotate the shoulder, one would rotate forwards and the other backwards. As it is now, it is also the standard rotation, if you symmetrize the bones.
In my opinion, it is therefore more beneficial to keep it that way so we are able to properly mirror bones and keyframes for future animations..

ipg-sig:
> We tried this out, but this only worked in the shoulders.

Thats a little bit strange. It should work for all mirrored bones (full arms, legs and eyes). I noticed that the Upper_Arm is named incorrectly on the right side, therefore the mirroring doesn´t work as expected, but works on all other bones for me.

> We also had a look at the Rigify default figure, there, also the orientations are the same for left an right.

For the legs it is possible to keep them in the same orientation (same for the eyes).
Flipping the keyframes works of course too, but could be more complex if someone wants to transform the bones inside the simulation. If we keep it like that, the right side is always mirrored with the orientation and we don´t have a mix. But I don´t have a strong opinion on that. So if you like, we can also discuss it in the group, if there are any preferences :)

ipg-sig:
@ClemensLinnhoff All requested changes should be implemented. If something is wrong, just let me know :)
I deletd the "Bones" directory and updated the work file in the sharepoint under Documents/03_Development/Work_Files.

Should be clearer now with the new bone rotations.

ClemensLinnhoff:
The bones and their directions look good now!
But there is a slight mismatch between the structure in the model and the structure in the documentation:

image

image

The "Armature_Human" is not in the structure. There it is called Skeleton. But it only contains the bones in the structure and not the meshes like in the model. Do we need to adapt the structure or the model @ipg-sig?

ipg-sig:
@ClemensLinnhoff We need to adapt the structure, so the meshes are a child of the Skeleton/Armature.

I think we should stay with "Armature" for Blender users. As far as I have read there are some import bugs from Blender to Unreal, which depend and deal with the term "Armature". If we rename it, it seems to create extra bones in the hierarchy,
But I am not sure if the postfix may create issues.

> Basically, the Armature Object in Blender becomes a bone, the root, when brought into Unreal, and that’s a problem if the rig already has its own root, so to prevent that they added a check that verifies if the first bone in the hierarchy is called “Armature”, and if it is, it then gets removed, and the hierarchy starts from the next bone over. MAYBE if you had a second root also called “Armature” it could work, I haven’t tested that, but this is the explanation for the behavior you’ve been seeing.

Source: https://forums.unrealengine.com/t/how-to-import-rigged-model-from-blender-to-unreal-engine-with-armature-root-bone/742089/5
.
@MatthiasThDs Do you have any experience with it? Does it still work this way?

ClemensLinnhoff:
I renames Armature_Human to Armature, so it fits to the defined structure.
This should be ready to merge now.
@AsamDiegoSanchez can you approve this so we can merge?

AsamDiegoSanchez (approved):
General revision done

PR #166, opened 2024-10-31
Add main introduction

### LINKED ISSUE #164: Add main introduction page

### LINKED ISSUE #168: Sharing the same 3D data file (FBX/gltf/USD) among multiple asset files.
In the current specification, the asset file and the 3D shape file are "implicit references" that refer to files with the same file name.
This means that one 3D data file can be associated with only one asset file.
Even if there are parked vehicles lined up on the map and vehicles of the same type (same 3D shape) but only different colors are lined up, it is necessary to prepare a 3D shape file that differs by the number of vehicles. (3D shape data cannot be shared and used)
This means that the 3D data cannot be shared.
The diagram below illustrates the issue. 
The asset files "Cube.xoma" and "Cube2.xoma" are based on the exact same shape "Cube.fbx," but "Cube.fbx" cannot be shared.
(You would need to copy "Cube.fbx" to "Cube2.fbx" for separate use.)
image

The solution to this issue is to move away from the "implicit reference" between the asset file and the 3D geometry file and replace it with an "explicit reference".
Specifically, in each asset file, specify the referenced 3D shape file.

However, even if we change to "explicit referencing," there remains another issue.
In the diagram above, the sphere within the 3D data file has a material name called "Material_Sphere," and its content is assigned with an actual material file in the Material Mapping file.
If this 3D Data file is shared with multiple Asset files, the contents of the "Material_Sphere" (referenced) must be specified for each Asset file, but since the assignment is determined in the Material Mapping file, in this case, the Material Mapping file will not be able to be shared.
The purpose of the Material Mapping file is to share material information among multiple asset files, so not being able to achieve this defeats the original purpose.
One possible solution is to reassign the referenced material name within the asset file (or prepare a separate Material Mapping file).
The diagram below illustrates the proposed solution, with the addition of the corresponding parts (highlighted in purple).
image

ClemensLinnhoff:
## Describe your changes

Add main introduction section.

Fixes #164, fixes #168

### DIFF
--- a/content/general-docs/introduction.adoc
+++ b/content/general-docs/introduction.adoc
@@ -46,10 +46,19 @@ The first interaction with an ASAM OpenMATERIAL-compliant 3D model is through th
 
 The *asset file* is a JSON file with the file extension .xoma (ASAM OpenMATERIAL Asset).
 A detailed definition of the https://github.com/asam-ev/OpenMATERIAL/blob/main/schemas/asset_schema.json[asset file JSON schema] is given in xref:geometry/asset-schema.adoc[].
+The asset file shall have the same name as a corresponding 3D data file in glTF, FBX or USD format.
+To facilitate instancing, multiple asset file may be linked to a single 3D model file by adding an index separated by a dot (.) as a suffix to the asset file name.
 It contains metadata, such as description, unique identifier, version information, copyright details, and so on.
 It also includes a material texture assignment table, allowing dedicated material mapping textures to be assigned to materials within 3D data files.
 This enables comprehensive material properties to be mapped to the geometry at a texture level.
-Each texel of the texture contains a 24 Bit code (8 Bit values in the rba channels) which is mapped to a material property file in a separate material mapping file, linked in the asset file with a URI.
+Each texel of the texture contains a 32 Bit code (8 Bit values in the rgba channels) which is mapped to a material property file in a separate material mapping file, linked in the asset file with a URI.
+For the mapping currently only the rba channels are used (first 24 Bit).
+The last channel is reserved for future use.
+
+The *3D data file* is a standard 3D model file in glTF, FBX or USD format.
+Multiple formats can exist in parallel, so one asset file can have multiple 3D data files with different formats with the same name as the asset file.
+However, to be standard compliant, a simulation environment must support at least one of the named 3D data formats.
+More information about the file formats is given in is given in xref:geometry/file-format-support.adoc[].
 
 The *material mapping file* is separated from the asset file to facilitate reusability.
 Multiple assets can link to the same material mapping, allowing simulation systems to have a unified mapping table for all assets. 


ClemensLinnhoff:
@FKlopfer could you do an editorial review? Especially check if I did the references to other sections correct, e.g. xref:geometry/asset-schema.adoc[]. This should display the section number and link to it in the documentation.

content/general-docs/introduction.adoc
```diff
@@ -3,11 +3,81 @@
...
+As a result, previous simulations were primarily used during the early stages of development, where precise physical modeling was not necessarily required. 
+
+With increased computing power and streamlined simulation techniques, virtual testing plays a more powerful role at all stages of system development.
+This correlates with the growing need to accurately represent the real world in simulations, so-called digital twins, especially as the need for physical sensor simulation increases. 
+
+One fundamental step to describing the real world virtually is a full physical description of all objects, including their material properties and their 3D geometries.
+
+It is extremely important to standardize the representation of data and attributes, such as scenarios, road data, and asset data, necessary for AD/ADAS simulation in order to promote the circulation of these data and to obtain consistent results among different systems. 
+
+Standardization activities have been conducted for scenarios and road data using OpenSCENARIO, OpenDRIVE, OpenCRG, and other standards.
```
AsamDiegoSanchez, line 27: we need to use the name "ASAM" at the beginning: 
ASAM OpenSCENARIO; ASAM OpenDRIVE,..

content/general-docs/introduction.adoc
```diff
@@ -3,11 +3,81 @@
...
+ASAM OpenMATERIAL closes the gap to advance standardization by providing standardizes definitions of material properties and 3D geometry.
```
AsamDiegoSanchez, line 28: We have to think if we want to stick with this name, and if so, write somewhere that in this document that we will use ASAM OpenMATERIAL in short, instead of ASAM OpenMATERIAL & 3D model structure

content/general-docs/introduction.adoc
```diff
@@ -3,11 +3,81 @@
...
+A detailed definition of the https://github.com/asam-ev/OpenMATERIAL/blob/main/schemas/asset_schema.json[asset file JSON schema] is given in xref:geometry/asset-schema.adoc[].
+It contains metadata, such as description, unique identifier, version information, copyright, etc.
+The asset files also contains a material texture assignment table.
+In this table, dedicated material mapping textures can be assigned to materials within a 3D data files.
+This provides to possibility to map comprehensive material properties to the geometry on a texture level.
+Each texel of the texture contains a 24 Bit code (8 Bit values in the rba channels).
+This code is mapped to a material property file in separate material mapping file, described in the next section.
+The material mapping file is linked in the asset file with a URI.
+
+The *material mapping file* is separated from the asset file to facilitate resuability.
```
AsamDiegoSanchez, line 59: Typo: "reusability"

content/general-docs/introduction.adoc
```diff
@@ -3,11 +3,81 @@
...
+image::openmaterial_overview.png[,1100]
+
+The first touching point with an OpenMATERIAL compliant 3D model is the asset file.
+
+The *asset file* is a JSON file with the file extension .xoma (OpenMATERIAL Asset).
+A detailed definition of the https://github.com/asam-ev/OpenMATERIAL/blob/main/schemas/asset_schema.json[asset file JSON schema] is given in xref:geometry/asset-schema.adoc[].
+It contains metadata, such as description, unique identifier, version information, copyright, etc.
+The asset files also contains a material texture assignment table.
+In this table, dedicated material mapping textures can be assigned to materials within a 3D data files.
+This provides to possibility to map comprehensive material properties to the geometry on a texture level.
```
AsamDiegoSanchez, line 54: Typo: "This provides "the" possibility...

AsamDiegoSanchez (changes requested):
I´ve found some minor typos, but overall looks good

content/general-docs/introduction.adoc
ClemensLinnhoff: Thanks, I fixed it.

content/general-docs/introduction.adoc
ClemensLinnhoff: Thanks, I fixed it.

content/general-docs/introduction.adoc
ClemensLinnhoff: Thanks, I fixed it.

content/general-docs/introduction.adoc
ClemensLinnhoff: The question is, what the official name of the standard will be. ASAM OpenMATERIAL or ASAM OpenMATERIAL & 3D model structure. I would favour simply using ASAM OpenMATERIAL as it is shorter.

AsamDiegoSanchez: Agree, we can talk about it this week with the whole group.

AsamDiegoSanchez (approved):
Remaining question about the standard name will be discussed.

content/general-docs/introduction.adoc
```diff
@@ -3,11 +3,74 @@
...
+ASAM OpenMATERIAL advances standardization by providing standardized definitions of material properties and 3D geometry.
+
+=== The two parts of ASAM OpenMATERIAL
+
+ASAM OpenMATERIAL consists of two main parts:
+
+* *Material*: This part includes definitions and file formats for storing and exchanging material properties. These properties can be physical, such as surface roughness, permittivity, and index of refraction, or measured data, like wavelength and angle-dependent reflectance values.
+* *Geometry*: This part contains node hierarchies for different object classes. These structures use a uniform naming scheme and coordinate definitions to enable the exchange of 3D models between various simulation systems.
+
+=== How everything fits together
```
FKlopfer, line 36: Maybe we can adapt this subtitle into something like "Integrating components seamlessly" or "Harmonizing components of ASAM OpenMATERIAL"?

ClemensLinnhoff: Or maybe "The structure of OpenMATERIAL"?
Because this section is mainly about all the file formats or OpenMATERIAL and how they are linked.
Or should we just call it "File formats"?

FKlopfer: Or we can combine this: "Structure and file formats of ASAM OpenMATERIAL"?

ClemensLinnhoff: I like that. But let's shorten it to "Structure and file formats".
The previous paragraph also describes the structure. But I would still like to keep it separate, because it has been a topic of discussion, that the two parts of materials and geometry shall be clearly separated, so they can be referenced separately. I think the extra section about this makes it more clear.

FKlopfer:
I also checked the links, they look good (screenshot from my fork): 
image

FKlopfer (approved):

content/general-docs/introduction.adoc
ClemensLinnhoff: - Generally a shorter name would be prefered. We could officially name the standard "**ASAM OpenMATERIAL & 3D model structure**" but everybody would just call it "OpenMATERIAL" as a nickname.
- Technically, something like **OpenASSET** or **Open3DASSET** would be more correct, because asset refers to the entire 3D model including its materials. And this definition would enable future additions, like animations etc.
- A workaround could be naming it **OpenMATERIAL+**, **OpenMATERIALx** or **OpenMATERIAL3D** to make clear that there is more to the standard than "just" materials.
- We could have a shorter name that still facilitates both parts, e.g. **OpenMATERIAL-ASSET**, **OpenMATERIAL-GEOMETRY**
- We could just stick with **OpenMATERIAL** and people will learn over time that it also contains geometry.

To be discussed further in the geometry group @LudwigFriedmann. Then we should put it to a vote.

ClemensLinnhoff: Let's move this discussion to the corresponding issue #169 and fix this in another PR.

Issue #133, opened 2024-10-09
Complete chapter Human structure (7.3.4)


LudwigFriedmann:
**Is your feature request related to a problem? Please describe.**
(TBD) Currently, the chapter and its subchapters are lacking text and illustrations. Even though the human structure has slightly changed, https://github.com/asam-ev/OpenMATERIAL/blob/OpenMATERIAL_Legacy/model_structure/Pedestrian_Structure.md  and the Miro board can serve as sources of information.

**Describe the solution you'd like**
Text and illustrations are added to the chapter.

**Describe alternatives you've considered**
-

**Additional context**
-

ClemensLinnhoff:
We might need to change the structure a bit. The mesh(es) that are moved by the armature need to be part of the armature hierarchy.
So we need something like

- Root
  - Armature
    - Skeleton
      - Root_Bone
        - Hip
        - ...
    - Body
    - Hair
    - Clothing
  - Accessory
  
  I will upload an example model shortly on this branch: https://github.com/asam-ev/OpenMATERIAL/tree/157-provide-example-for-a-human-asset-analog-to-vehicle-example

PR #181, opened 2024-11-19
133 complete chapter human structure 734

ipg-sig:
## Describe your changes

Added content to the human structure.

Issue: Complete chapter Human structure (7.3.4) #133 

@ClemensLinnhoff @LudwigFriedmann @MatthiasThDs @KimuraDIVP Please feel free to review and comment.
Work files can be found in the sharepoint under Documents/03_Development/Work_Files

The following adjustments have been made to the example model and structure according to the Miroboard:

The second detailed LOD level was set as the default, but simplified

- added bones for foot and toe
- added bones for hand and thumb
- added bones for hip
- added bones for eyes
- added groups for root, accessories, clothing, hair, body
- renamed some bones
- adjusted axis orientation
- adjusted bone orientation, so mirroring would be possible to use for animationg the character
- Added hip bones as seperate bones to stay compatible between (future) LOD levels. Therefore the bones can all have an unique name and can be used for (future) animations without breaking them. Same for the hand and spine.

Full_Hand ← Hand → Hand (with Upper_Fingers and Lower_Fingers)
Full_Foot ← Foot → Foot (with Upper_Toes and Lower_Toes)


content/geometry/object-human/human-model-structure.adoc
```diff
@@ -2,1213 +2,997 @@
...
+          |Foot_Right
+            |Full_Toes_Right
+  |Grp_Accessories
+    |Accessorie
+  |Grp_Body
+    |Human
+  |Grp_Clothing
+    |Vest
+  |Grp_Hair
+    |Hair
```
ClemensLinnhoff, line 44: Links to the sections are missing here. @FKlopfer you assist in adding them back in?

content/geometry/object-human/human-model-structure.adoc
```diff
@@ -2,1213 +2,997 @@
...
-| ...
-|===
-
-
-Add a figure.
-
-.<Figure caption>
-//image::images/Vehicle_Structure_Door_Coord_Frame.svg[width=70%, scalewidth=10cm]
+This group is used as a parent for all following nodes. It can be used to place the whole asset and select the complete node hierarchy at once.
+The origin of the Root group is set at the world origin.
```
ClemensLinnhoff, line 51: I would not call it "world origin" as it is the origin of the object, i.e. center of bounding box on the ground, correct?
Same for the other Grp nodes.

ClemensLinnhoff:
In the example, there is the Armature as a hierarchy element between Grp_Root and the other Groups (Body, Hair etc.), like this:

- Grp_Root
  - Armature
    - Armature_Human
      - Root
        - Lower_spint
        - ...
    - Grp_Accessories
    - Grp_Body
    - ...
    
I think it needs to be that way, because every mesh that is affected by the skeleton needs to be part of the armature, as far as I know. So do we have to put this in the structure here in the documentation?

content/geometry/object-human/human-model-structure.adoc
FKlopfer: Added links to structure overview. Some chapters exist only in the overview yet, they need to be created: Skeleton, Human, Vest, Hair, Accessorie.

The order of the chapters also needs to be adjusted when everything is finished.

content/geometry/object-human/human-model-structure.adoc
ipg-sig: I would say if it is done right, the world origin should equal the origin of the object (middle of the bounding box on the ground), so that the transforms are set to 0 0 0. But I can change it here or state, that those should be the same in the end, if it is set up correctly.

ipg-sig: Good point. I will change it.

ClemensLinnhoff (approved):

ClemensLinnhoff:
@FKlopfer This is ready for an editorial review.

ClemensLinnhoff:
We just noticed: shoulder right is missing @ipg-sig

mettelilienthal:
We have a few questions regarding consistency between the vehicle, road, and human structure:

- Title: For the vehicle and road structure, the titles name the indices; while for the human structure, simply the bone itself is the title (Grp_Door_Bottom_<door_bottom_idx> vs. Upper_Spine). Which title naming convention should we work with?

- Origin definition: For the vehicle and road structure, the origin is defined in the tables along the axes; while for the human structure, the origin is defined in the text above the tables. Should we move the origin definitions to the tables?

- Definition of groups: In the vehicle structure, we introduced each group's definition with "This group contains..."; while in the human structure, the group is named explicitely ("The `Lower_Spine` bone represents..."). Should we rework the vehicle structure accordingly? This might help searchability as naming the group is more concrete.

- Coordinate systems: For the vehicle structure, we never defined how the local coordinate system for the vehicle as a whole is placed in relation to the world coordinate system. Should we add that and if so, how?

ClemensLinnhoff:
- Title/Groups: There is a difference between a bone and a group. A group is an empty node grouping multiple meshes or further sub-groups. A bone is a single element in an armature/skeleton. Bones only exist once, so they don't have an index.
- Origin definition: You are right, let's harmonize the tables and put the origin in the tables here.
- Coordinate systems: This is also defined for the vehicle: "The origin of the Root node is the center of the vehicle’s bounding box projected to the ground, including all vehicle parts in their default positions". It should actually always be the center of the bounding box ob the object projected to the ground for all object classes.

mettelilienthal:
Thanks for the feedback!

Another question regarding the tables: The tables for the objects that can move (Hair, Clothing, ...) differ because they list the transformation. Would it be possible to extract that information from the table, rather listing it in the definition to ensure consistency? We would then write something like "This object typically moves in the simulation and therefore has its own transforms."

Also, what kinds of transformations are we talking about? Is it only rotation (as mentioned in 7.2.2 Coordinate systems) or also scaling etc.? Maybe we should specify this?

ClemensLinnhoff:
Hair, Clothing etc. actually don't move independently in the simulation. Their coordinate system stays in the center of the bounding box on the ground, so the object coordinate system.

mettelilienthal:
Could you then please explain why these objects have a transformation listed in the table? 
image

ClemensLinnhoff:
> Could you then please explain why these objects have a transformation listed in the table? 

In my opinion they could be removed here. @ipg-sig @LudwigFriedmann what do you think?

ipg-sig:
> > Could you then please explain why these objects have a transformation listed in the table?
> 
> In my opinion they could be removed here. @ipg-sig @LudwigFriedmann what do you think?

@ClemensLinnhoff I think we can remove it, if it is clear, that they shouldn´t have any transformations (Location & Rotation at 0, Scale at 1), to be able to properly use them.

> The coordinate definitions of the bones are not clear all the time. The y-axis should always be along the bone, right? And the x-axis always forward? Then I would define it that way instead of using up, down, side etc. Might make it more clear for bones that are at an angle in the default position.

If we want to keep the ability to mirror it properly, it is not always forwards. One side is forwards, the other one backwards. I think images of the bone rotation makes it more clear. But we can also change it or call it otherwise. What or how would you prefer to call it to keep it consitent and clear for the reader?

ClemensLinnhoff:
About the bones, the y-axis is consistently along the bones, right? So we could define it that way.
The x-axis in the example is sometimes pointing forward, sometimes backwards. If we want to keep it that way, we need to adjust the specification here to match that.

content/geometry/object-human/human-structure.adoc
```diff
@@ -0,0 +1,861 @@
...
+
+| *z-axis*
+| Pointing upwards
+|===
+
+NOTE: Please add a figure and description.
+
+== Lower_Spine
+
+The `Lower_Spine` bone represents the lowest parts and bones of the spine, that is, the Coccyx and Sacrum spine bones of the human skeleton.
```
mettelilienthal, line 191: @ClemensLinnhoff please check whether this definition structure works for you :)

content/geometry/object-human/human-structure.adoc
```diff
@@ -0,0 +1,861 @@
...
+
+| *z-axis*
+| Pointing sidewards
+|===
+
+NOTE: Please add a figure and description.
+
+== Lower_Arm_Left
+
+The `Lower_Arm_Left` bone represents the lower part of the left arm, that is, the left elbow and Radius and Ulna of the human skeleton.
```
mettelilienthal, line 416: Is it correct to add the bones here? Please also check for the legs accordingly.

content/geometry/object-human/human-general.adoc
```diff
@@ -1,3 +1,7 @@
 = General
 
-//Add general information about the human structure
\ No newline at end of file
+A human is a bipedal being that can be represented by an object in 3D geometry according to the human body's physical representation.
```
mettelilienthal, line 3: @ACMackenthun please review :)

ACMackenthun, line 3:
```suggestion
A human is a bipedal being that can be represented by an object in 3D geometry.
```

content/geometry/object-human/human-general.adoc
```diff
@@ -1,3 +1,7 @@
...
+
+The human 3D geometry is generically structured into the node structure and skeleton (also often called armature) defined in the <<Structure overview>>. The structure starts with a Root node as a parent group for all other nodes in the structure. Splitting different object parts, such as clothing, hair, and accessories enables the animation and exchange of parts. Not all nodes have to be present in every human 3D model. Additional object parts may be added to facilitate and better visualize additional use cases. They should adhere to the same structure principle.
```
ACMackenthun, line 5:
```suggestion
The human 3D geometry is generically structured into the node structure and skeleton (also often called armature) defined in the <<Structure overview>>. The structure starts with a Root node as a parent group for all other nodes in the structure. Splitting the object into different parts, such as clothing, hair, and accessories, enables the animation and exchange of parts. Not all nodes have to be present in every human 3D model. Custom object parts may be added to facilitate and better visualize additional use cases, but all custom parts shall adhere to the specified structure principle.
```

content/geometry/object-human/human-general.adoc
```diff
@@ -1,3 +1,7 @@
...
+
+The human 3D geometry is orientated at x as the axis facing forwards, y as the axis facing sidewards, and z as the axis facing upwards. The default pose is the A-pose. The naming convention makes it possible to mirror one side of the skeleton to the other. The bone orientation allows (future) animations to be mirrored an both sides.
```
ACMackenthun, line 7:
```suggestion
In the coordinate system for human 3D geometry, the x-axis points forwards, the y-axis points sidewards, and the z-axis points upwards. The default pose is the A-pose. The naming convention makes it possible to mirror one side of the skeleton to the other. The bone orientation allows (future) animations to be mirrored an both sides.
```

ACMackenthun:
I noticed that today's force push reverted all of our linguistic changes - was there a specific reason for that?

ipg-sig:
@ACMackenthun No, that was not supposed to happen. I messed something up with git and I am already trying to fix it at the moment :(
I still should have all linguistig changes locally, so nothing should be lost.

ACMackenthun:
@ipg-sig: No worries then, just wanted to check that we haven't done anything wrong :) Let us know if we can do anything to help

ipg-sig:
@ClemensLinnhoff All requested changes should be implemented. If something is wrong, just let me know :)

ClemensLinnhoff (commented):
Looks very good!
I just have one question @ipg-sig: Should we replace "Pointing upwards" for the y-xis of the bones with "Pointing along the bone"?
Because in the default A-pose it is not really upwards for all the bones and might be misleading. What do you think?

ACMackenthun (approved):

FKlopfer (approved):

ipg-sig:
@ClemensLinnhoff You mean in the tables? I think upwards, sidewards and forwards is shorter and keeps the reading flow. But we could add an sentence at the beginning which states, that the y-axis is always following the bone direction, regardless if it is sidewards, upwards or forwards.

ClemensLinnhoff:
> @ClemensLinnhoff You mean in the tables? I think upwards, sidewards and forwards is shorter and keeps the reading flow. But we could add an sentence at the beginning which states, that the y-axis is always following the bone direction, regardless if it is sidewards, upwards or forwards.

Yes, I think this would be good.

content/geometry/object-human/human-general.adoc
```diff
@@ -1,3 +1,13 @@
...
\ No newline at end of file
+A human is a bipedal being that can be represented by an object in 3D geometry.
+
+The human 3D geometry is generically structured into the node structure and skeleton (also often called armature) defined in the xref:../geometry/object-human/human-index.adoc#_model_structure[Model structure].
+The structure starts with a Root node as a parent group for all other nodes in the structure.
+Splitting the object into different parts, such as clothing, hair, and accessories, enables animation and exchange of parts.
+Not all nodes have to be present in every human 3D model.
+Custom object parts may be added to facilitate and better visualize additional use cases, but all custom parts shall adhere to the specified structure principle.
+
+In the coordinate system for human 3D geometry, the x-axis points forwards, the y-axis points sidewards, and the z-axis points upwards.
```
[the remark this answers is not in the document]
mettelilienthal: @ClemensLinnhoff @ipg-sig is this what you wanted?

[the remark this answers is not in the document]
ClemensLinnhoff: I think "in the default pose" can be removed. The y-axis always points in the direction of the bone, in all poses.
So this basically only applies to the x- and z-axis that they are forwards and sideways in the default pose.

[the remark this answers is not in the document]
ipg-sig: This only specifies the general coordinate system of the 3D assets for humans, not the orientation of the bones. Therefore I would leave this sentence as it is and add a new one, which specifies the bone direction.

+ In the coordinate system for human 3D geometry, the x-axis points forwards, the y-axis points sidewards, and the z-axis points upwards.
+ The default pose is the A-pose.
+ The naming convention makes it possible to mirror one side of the skeleton to the other.
+ The bones follow a specific bone orientation. The y-axis of a bone is always following the bone direction, regardless if it is pointing sidewards, upwards or forwards. The x-axis is mostly facing forwards.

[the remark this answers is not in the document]
ClemensLinnhoff: That sounds better, thanks!

content/geometry/object-human/human-general.adoc
```diff
@@ -1,3 +1,13 @@
...
+The default pose is the A-pose.
+The naming convention makes it possible to mirror one side of the skeleton to the other.
```
mettelilienthal, line 13:
```suggestion
The naming convention makes it possible to mirror one side of the skeleton to the other.
The bones follow a specific bone orientation: The y-axis of a bone always follows the bone direction, regardless of whether it points sidewards, upwards, or forwards. The x-axis usually faces forwards accordingly.
```

content/geometry/object-human/human-general.adoc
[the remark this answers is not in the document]
mettelilienthal: I've updated the file accordingly :)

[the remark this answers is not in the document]
ipg-sig: Thank you! :)

ClemensLinnhoff:
So now, we still need to change the structure to fit to the example implementation, right @ipg-sig ?

ipg-sig:
@ClemensLinnhoff Yes, that´s right. I am just waiting if we should keep the name "Skeleton" or if we should change it to "Armature", because of the Blender to Unreal import behavior. I am not sure if this is still needed with UE5. But iff nothing speaks against it, I would change it to "Armature", so we are safe in both cases.

ipg-sig:
@ClemensLinnhoff Done :)

Issue #182, opened 2024-11-20
Support for extensions.


Norihito-Hiruma:
We would like to keep  information preparing for the future function expansion like (1)additional consideration information for V2.0 and later, (2)original information required for each application, and (3) information for  the additional sensors.
The glTF already have a very good format for the extensions mentioned above. 
We suggest that OpenMATERIAL also utilize glTF as a reference and add "extensionsUsed", "extensions", and "extras" items. 
https://github.com/KhronosGroup/glTF/tree/main/specification/1.0#specifying-extensions
https://github.com/KhronosGroup/glTF/tree/main/extensions

![OpenMATERIAL_機能拡張追加2](https://github.com/user-attachments/assets/2eedb143-3563-4522-93fc-1f8989a3df8e)

ClemensLinnhoff:
Would it be okay if we move this to the next OpenMATERIAL project/version? We only have about 2 weeks left to complete the content of this project wand we still have a lot of open topics.

ClemensLinnhoff:
A simpler implementation of this could be added already in the current version of OpenMATERIAL: Just adding a customProperties field to the xomp file.

ClemensLinnhoff:
We should also add `customProperties`  to the asset file to allow custom extensions there, too.

Norihito-Hiruma:
I added customProperties to the asset file, and added it to the Pull request #199.

PR #199, opened 2024-11-26
Add customProperties field to material file

Norihito-Hiruma:
modified:   schemas/material_schema.json

## Describe your changes

In response to the result of the discussion in issue #182, customProperties have been added to the material file schema.

**Some questions to ask**:

I have added an item with reference to glTF's extensions, so please review it.

Related issue is #182.


schemas/asset_schema.json
```diff
@@ -181,6 +181,15 @@
...
       }
+    },
+    "customProperties" : {
+      "type" : "array",
+      "description" : "Names of custom properties used somewhere in this asset.",
+      "items" : {
+        "type" : "string",
+      },
+      "uniqueItems" : true,
+      "default" : []
```
ClemensLinnhoff, line 192:
```suggestion
    "customProperties": {
      "type": "object",
      "description" : "Non-standardized asset properties for custom tools or tool chains.",
      "propertyValues": {
        "type": "string"
      }
    }
```

I would propose to do it like this, as this allows adding any number of key-value pairs as properties, like in this example:

```JSON
"customProperties": {
  "test": "my first custom property",
  "test2": "my second custom property"
}
````

File: schemas/material_schema.json [hunk 2]
```diff
@@ -185,6 +185,18 @@
...
+          "description" : "Custom properties in this material.",
+          "items" : {
+            "type" : "object",
+            "properties" : {
+            },
+            "additionalProperties" : {
+              "type" : "object"
+            }
+          }
         }
```
ClemensLinnhoff, line 200:
```suggestion
    "customProperties": {
      "type": "object",
      "description" : "Non-standardized material properties for custom tools or tool chains.",
      "propertyValues": {
        "type": "string"
      }
    }
```

Slight adjustment to be harmonized with the definition in the asset file.
```

ClemensLinnhoff (changes requested):
I propose slight adjustments to the schema definitions.
Would this be okay for you @Norihito-Hiruma?

schemas/material_schema.json
Norihito-Hiruma: Thank you, Clemens-san, for checking the content and suggesting corrections.

Excuse me, but what is the definition of json schema propertyValues?
No propertyValues were found in the json schema specification.
[JSON Schema Specification](https://json-schema.org/specification)

Anyway,
- In the asset file, declare a descriptive extension name.
- In the material file, you can set multiple detailed parameters and their values included in the declared extension.
This parameter value can be set to either a string or a number.

Is it correct to understand that?
If my understanding is not wrong, I agree with your suggestion.

ClemensLinnhoff: "propertyValues" is basically the same as the "properties" field in your proposal. It is just a name. We can also change it back to "properties".

> In the asset file, declare a descriptive extension name

In the meeting I proposed that we make it more generic. So you can use the custom properties to describe what kind of extension you use in the materials, but you can also use it for other custom properties directly in the asset file.

Norihito-Hiruma: Clemens-san, Thank you for the explanation.
It seems that I missed what you explained during the meeting. I'm sorry.
I had understand your opinion.
I agree with your amendment suggestion.

ClemensLinnhoff (approved):

PR #217, opened 2024-12-04
Added image for human skeleton

ACMackenthun:
## Describe your changes

Add a description of the changes proposed in the pull request.

**Some questions to ask**:

- What is the change?
- What does it fix?
- Is this a bug fix or a feature? Does it break any existing functionality or force me to update to a new version?
- How has it been tested?

Add a reference to a related issue.

## Mention a member

Add @mentions of the person or team responsible for reviewing the proposed changes.


content/_images/fig_human-structure.svg
ACMackenthun, line 1: I added an image of the human skeleton to the human-general.adoc file. I decided to add the labels for Full_Thumb_Left, Full_Fingers_Left, and Full_Toes_Left to illustrate the different strategies for hands and feet.

However, with this approach I had to leave out the labels for Foot_Left and Hand_Left. Expecially the label for Foot_Left does not fit, because the origin of Full_Toes_Left is in the middle of the foot geometry and the line between the ankle point and the origin of the toes does not represent the full length of Foot_Left.

@ipg-jsc, @ClemensLinnhoff: Do you prefer the image with fingers and toes or should we focus on the more abstract hand and foot level?

ClemensLinnhoff: I think it might make sense to have two images of the pedestrian. One with the lower level of detail, similar to the [image in OSI](https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1MovingObject_1_1PedestrianAttributes_1_1Bone.html#ac25b7db490a6e225ba636df8e01fda88). And one with the higher level of detail with the fingers, toes and eyes.

About the coordinate system: This needs to be moved to the center of the bounding box projected to the ground. The orientation is correct.

And I think it would be a good idea to have a 3rd image, with just one generic bone and the bone coordinate system. Something like this (with a visible y-axis along the bone and of course in the style of all other images):
image

ipg-sig: I think Clemens proposal would work best.

We changed some of the bones names, therefore the names in the image also need to be adjusted accordingly. I also noticed that the root bone is missing.

mettelilienthal:
We have a question regarding the shoulder: It is defined as "the upper part of the left shoulder, that is, the interaction between the clavicle bone and the humerus head of the human skeleton."
However, we do not have a lower part of the shoulder currently defined - should we add something?

Also, we were wondering whether the clavicle bone should be defined.

And, regarding the human structure figure: What do the dotted lines represent? Is that the clavicle or the interaction of the shoulder? Or would one single line from neck to upper arm be enough representation?
image

ClemensLinnhoff:
> We have a question regarding the shoulder: It is defined as "the upper part of the left shoulder, that is, the interaction between the clavicle bone and the humerus head of the human skeleton." However, we do not have a lower part of the shoulder currently defined - should we add something?
> 
> Also, we were wondering whether the clavicle bone should be defined.

We cannot add more boned in-between, because the OpenMATERIAL definitions need to fit to the OSI definitions.

ClemensLinnhoff:
> And, regarding the human structure figure: What do the dotted lines represent? Is that the clavicle or the interaction of the shoulder? Or would one single line from neck to upper arm be enough representation?

As far as I can see it, the shoulder bones do not have to be directly attached to the end of the upper spine. The can have an offset:
image

mettelilienthal:
Thanks for the information! Then I would mention the offset in the shoulder definition to make it clear to everyone.

mettelilienthal:
image
Could you please confirm that the clavicle bone is still correct in the definition? Even though we are talking about the upper spine in the offset?

ClemensLinnhoff:
> image Could you please confirm that the clavicle bone is still correct in the definition? Even though we are talking about the upper spine in the offset?

@ipg-sig could you confirm that?

ipg-sig:
@mettelilienthal @ClemensLinnhoff The definition of the shoulder bone is still correct with the mention of the claivcle bone, yes. It only describes the movement/ interaction of the real body part.

The other sentence with the offset states, that the bone doesn´t need to be directly connected to the upper spine bone. So it shouldn´t be confusing for the reader :)

mettelilienthal:
Thanks for the information! Then I'll update this PR with the correct files in the following days :)

ClemensLinnhoff:
I noticed, that Shoulder_Right is missing in the Structure overview. The section exists, it is just missing from the overview. 
@ACMackenthun could you simply add that here? Then we don't have to create a separate issue.

mettelilienthal:
@ClemensLinnhoff could you please check whether this finalises the PR? :)

ClemensLinnhoff (changes requested):
Please adjust the blue coordinate frame. Its axes should be parallel to the corresponding edges of the bounding box.

And the Root bone is a bit hard to see, because the z-axis is hiding its start. Can you maybe move the black dot and line of the bone in front of the z-axis? And I would put the label Root over the line, similar to Lower_Spine.

Also, the end point of the Hip / Beginning of the Lower_Spine is missing. See image in OSI for reference: https://opensimulationinterface.github.io/osi-antora-generator/asamosi/latest/gen/structosi3_1_1MovingObject_1_1PedestrianAttributes_1_1Bone.html#ac25b7db490a6e225ba636df8e01fda88

mettelilienthal:
Just implemented your feedback - please review again :)

ClemensLinnhoff:
The beginning of the root bone needs to be at the origin.

ClemensLinnhoff (approved):

PR #233, opened 2024-12-12
143 required fields in the material schema

### LINKED ISSUE #143: Required fields in the material schema
#### Problem description
In the current version of the material schema, practically all fields are required. I could not find any materials properties which are marked optional. 

see https://github.com/asam-ev/OpenMATERIAL/blob/main/schemas/material_schema.json 

#### Question
My question is therefore: do we require anyone who contributes a material definition to know all the materials parameters for a sample that should be included?

#### Additional context
We had a discussion in the last meeting, that it is essential for the user/simulation engineer to be able to have a full set of properties of a material. 

However, from a database perspective, often not all parameters of a material have been measured or are known. 
In particular the elasticity or deformation properties might not be known for a database of reflection measurement samples. Another example are surface roughness, which is a parameter that applies across multiple scales but is often hard to measure with precision. 

#### Issue
I think it might be hard to add new material definitions in this standard if all parameters are required and need to be known. This could hinder a class of users (who are only interested in specific properties) to apply the standard in their implementations.

ClemensLinnhoff:
## Describe your changes

- Change required fields
- Harmonize field descriptions

Fixes #143 

### DIFF
--- a/examples/example_mapping.xomm
+++ b/examples/example_mapping.xomm
@@ -8,7 +8,7 @@
     "copyright": "(C) 2023-2024, Example Company",
     "license": "MPL-2.0",
     "author": "john.doe@asam.net",
-    "creationDate": "20240703T101728Z",
+    "creationDate": "20240703T101728Z"
   },
   "materialMapping": [
     ["Material_Sphere", "example_material.xomp", "white aluminum"],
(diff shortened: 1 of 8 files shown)


AsamDiegoSanchez (approved):
Reviewed.

Issue #71, opened 2024-08-22
Add vehicle class meta data


ClemensLinnhoff:
**Is your feature request related to a problem? Please describe.**
Some metadata depends on the object class.

**Describe the solution you'd like**
Add vehicle dependent metadata. Consider [vehicle definitions in OpenSCENARIO](https://publications.pages.asam.net/standards/ASAM_OpenSCENARIO/ASAM_OpenSCENARIO_XML/latest/generated/content/Vehicle.html).

ClemensLinnhoff:
It was only removed from the PR. So in my opinion we should still include it in the main asset file metadata.

MustafaTrian:
ok, by including we mean conditional including depending on the objectClass type , right?
Also, i noticed that in the main asset schema objectClass can be vehicle, human, road  netwok and "others" while under https://github.com/asam-ev/OpenMATERIAL/blob/main/content/geometry/object-classes.adoc
it is stated that the object can be only one of the three classes i.e. no "others" were mentioned

ClemensLinnhoff:
I would just make it an optional field. Conditions cannot be included in JSON schema files.
You are right. We should add a sentence to the documentation that you can have assets of other classes that are not defined (yet).

MustafaTrian:
What about https://json-schema.org/understanding-json-schema/reference/conditionals#dependentRequired ?

ClemensLinnhoff:
Interesting. I did not know about this. If this works, would be great!

PR #245, opened 2024-12-13
add vehicle class dependent metadata

MustafaTrian:
## Describe your changes

Add vehicle class dependent metadata to asset schema file following https://releases.asam.net/OpenSCENARIO/1.0.0/Model-Documentation/content/Vehicle.html

the original Axle object defined by https://releases.asam.net/OpenSCENARIO/1.0.0/Model-Documentation/content/Axles.html is a little bit misleading; the text description and properties describe additionalAxles
as list of (additional) axle that is an elemnt in the Axles list.
However, the XSD 1.0 Representation states that additional axles are simply axles and are direct sub elements in the Axles array forcing all of them to have the unified name "additionalAxle":
 <xsd:element name="AdditionalAxle" type="[Axle](https://releases.asam.net/OpenSCENARIO/1.0.0/Model-Documentation/content/Axle.html)" minOccurs="0" maxOccurs="unbounded"/>
 I followed the XSD Representation.

Some tests were made on the schema against input json example on https://www.lddgo.net/en/string/json-schema

[#71](https://github.com/asam-ev/OpenMATERIAL/issues/71)

## Mention a member

@ClemensLinnhoff 

- [] My changes generate no new warnings during the documentation generation.


ClemensLinnhoff:
Thank you @MustafaTrian for this!

I did some minor changes to the schema, e.g. I had to change the axles from an array to an object, because it has to have front and rear axle and only the additional axles is an array.

I also added some data to the vehicle example xoma file to test out the schema. It works quite well I think.

I am just not entirely sure if we really need the if-then condition. Because the vehicleClassData field is optional anyways. So I think we can just remove the if-then. This would make it much easier to parse the schema into the asciidoc documentation.
I will try this out in my next commit. We can revert it, if we decide we do need the if-then.

ClemensLinnhoff:
Using the definitions for the axles is a very good idea. I just need some time to adjust the python script generating the documentation, because it is currently not capable of handling definitions.

ClemensLinnhoff:
> Using the definitions for the axles is a very good idea. I just need some time to adjust the python script generating the documentation, because it is currently not capable of handling definitions.

This was actually a bit easier than I thought and I directly added it

ClemensLinnhoff:
In my opinion this is ready to be merged.
What do you think @LudwigFriedmann @ipg-sig @MatthiasThDs @DavidJRitter904 @KimuraDIVP ?

ClemensLinnhoff (approved):

LudwigFriedmann (approved):
Looks good to me!

Issue #254, opened 2024-12-18
Add more detailed definition in the BRDF description


ClemensLinnhoff:
**Is your feature request related to a problem? Please describe.**
There are different types of BRDFs

**Describe the solution you'd like**
Add definition of BRDF to description field in BRDF schema

Issue #236, opened 2024-12-12
Regarding a conversion from measured reflectance values with Lambertian reflector target to BRDF values


DavidJRitter904:
### Describe the problem
A separate discussion started within Issue https://github.com/asam-ev/OpenMATERIAL/issues/167 about using measured reflectance values referenced to a Lambertian reflector target within a BRDF. 

---
### Research done so far

_If anything of the information we gathered on this topic is missing, please include them with a comment below!_

#### BRDF definition

>  _the ratio of radiance in the outgoing direction to the incident irradiance [2]_

With the Assumptions below, we define a function $f_r$ that is the _basic quantity that characterizes (geometrically) the reflecting properties of a surface_ called "Bidirectional Reflectance Distribution Function", abbreviated BRDF:

$$ BRDF: f_r(\theta_i , \phi_i ; \theta_r , \phi_r) = \frac{dL_r(\theta_i , \phi_i ; \theta_r , \phi_i ; E_i)} {dE_i(\theta_i , \phi_i)} $$
$$ dE_i = L_i(\theta_i , \phi_i) d\Omega_i $$

Since for a given pair of directions, the BRDF $f_r$ is a concentration of reflectance (per steradian) it may take on any value from zero to infinity.

#### BRDF properties

Positivity: $f_r(\theta_i , \phi_i ; \theta_r , \phi_r) \ge 0$
Energy conservation: $\forall\omega_i; \int_{\omega_r} \int_{\omega_i} f_r(\theta_i , \phi_i ; \theta_r , \phi_r) cos(\theta_r)d\Omega_i d\Omega_r \le 1$
Helmholtz reciprocity: $f_r(\theta_i , \phi_i ; \theta_r , \phi_r) = f_r(\theta_r , \phi_r ; \theta_i , \phi_i)$

#### Reflectance Factor

Continue by using the definition of the **reflectance factor $R_F$** by Nicodemus et al.[1] which is the ratio of the radiant flux actually reflected by a sample surface to the radiant flux reflected by the ideal diffuse Lambertian reflector denoted as "id" ($\rho = 1$) in the same reflected-beam geometry, irradiated in exactly the same ways as the sample. 

Combining equation (12) for the reflected flux by the surface element $dA$ into the solid angle $\omega_r$

[...]

PR #259, opened 2024-12-18
Add more detailed definition for BRDF

DavidJRitter904:
## Describe your changes

Added a more detailed definition of the BRDF in the description field based on the references stated in #236 and [Hapke](https://books.google.at/books?id=3FNzaFuoXY0C&lpg=PP1&hl=de&pg=PP1#v=onepage&q&f=false).

#254 

## Mention a member

@KimuraDIVP @ClemensLinnhoff @lyndyRott @Norihito-Hiruma 
Ready for review

- [ ] I have performed a self-review of my code/documentation.
- [ ] My changes generate no new warnings during the documentation generation.

### DIFF
--- a/schemas/material_brdf_schema.json
+++ b/schemas/material_brdf_schema.json
@@ -69,7 +69,7 @@
         },
         "brdf": {
             "type": "object",
-            "description": "General information about the BRDF-lookup table. The BRDF is used as a basic quantity to characterize geometrically reflecting properties of a surface. BRDF-Definition: “The ratio of the radiance or power per unit solid-angle in the outgoing direction from a unit projected surface area to the collimated irradiance or power per unit projected surface area.” The BRDF has units of inverse steradians or solid-angle (sr-1) and could vary from 0 to infinity. It is symmetric in the incoming and outgoing direction, known as the Helmholtz reciprocity principle.",
+            "description": "General information about the BRDF-lookup table. The BRDF is used as a basic quantity to characterize geometrically the reflecting properties of a surface. BRDF-Definition: “The ratio of the radiance or power per unit solid-angle in the outgoing direction from a unit projected surface area to the collimated irradiance or power per unit projected surface area.” A physical plausible BRDF must obey the following three properties. (1) Positivity: The BRDF has units of inverse steradians (1/sr) and could vary from 0 to infinity. (2) Helmholtz reciprocity: It is symmetric in the incoming and outgoing direction. (3) Energy conservation: No more radiant energy is reflected from a point on the surface than incident radiant energy to that point. ",
             "properties": {
                 "wavelengths": {
                     "type": "array",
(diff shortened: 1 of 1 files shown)


lyndyRott (commented):
Well done. It makes sense to me to use this definition which refers to an angle pair (collimated ray "input") as opposed to the albedo from remote-sensing, where the incoming direction might not be known or arbitrary and thus the entire half-space is used.

TimoHinsemann:
Sounds very good, thank you very much! One thing I would like to add to the description - As far as I know, the part "...Could vary from 0 to infinity. It is symmetric in the incoming and outgoing direction, known as the Helmholtz reciprocity principle." is not part of the BRDF-Definition, but means that a BRDF is "physically plausible" (See Montes and Urena: https://digibug.ugr.es/bitstream/handle/10481/19751/rmontes_LSI-2012-001TR.pdf ; Page 3). So a BRDF is physically plausible, when it is non-negative (1), obeys symmetry (Helmholtz Reciprocity) (2) and no more radiant energy is reflected than the incident energy (3). 

Thats why I would prefer to add these 3 mentioned properties as "requirements" for a "physically plausible" BRDF, and not the BRDF-Definition. Also we could have a problem with the "symmetry" when it comes to measured BRDFs, which would be another reason to exclude the properties from the BRDF-Definition.

Norihito-Hiruma:
Thank you so much.
I think it's a very polite and good additional explanation.

lyndyRott:
> Sounds very good, thank you very much! One thing I would like to add to the description - As far as I know, the part "...Could vary from 0 to infinity. It is symmetric in the incoming and outgoing direction, known as the Helmholtz reciprocity principle." is not part of the BRDF-Definition, but means that a BRDF is "physically plausible" (See Montes and Urena: https://digibug.ugr.es/bitstream/handle/10481/19751/rmontes_LSI-2012-001TR.pdf ; Page 3). So a BRDF is physically plausible, when it is non-negative (1), obeys symmetry (Helmholtz Reciprocity) (2) and no more radiant energy is reflected than the incident energy (3).
> 
> Thats why I would prefer to add these 3 mentioned properties as "requirements" for a "physically plausible" BRDF, and not the BRDF-Definition. Also we could have a problem with the "symmetry" when it comes to measured BRDFs, which would be another reason to exclude the properties from the BRDF-Definition.

To provide the 3 properties makes sense, as it is done in various literature sources. Property (3) could also be shortened to the term "energy conservation", under which it is typically known.

ClemensLinnhoff:
Okay, @DavidJRitter904 would you add this addition from Timo? Then I think we are ready to merge.

DavidJRitter904:
Thank you for your feedback!
I have added the three properties for a physical plausible BRDF to the description. 
The part with the variation of of the units between 0 and infinity is added as there can be high radiance in exit direction due to arbitrary small irradiance. 

I think now we have a very elaborate definition.

ClemensLinnhoff (approved):

Issue #263, opened 2024-12-19
Human structure needs some fixes


LudwigFriedmann:
**Describe the bug**

1. The human structure table features very specific node names (e.g. Accessories_Wristband).
2. The term accessories should be elaborated. It should be made clear that it covers former Grp_Equipment_Dynamic (If equipment is very specific for a human asset, rather small, not self-contained or animated, it can be part of the human asset.) as individual meshes.
3. Notes from miro board should be removed.
4. As bones create local coordinate frames (transforms), those should be indicated in the structure and headlines.

**To Reproduce**
1. See human structure documentation

**Expected behavior**

1. Rename to more generic names (e.g. Accessories_<name>), consistent with the subchapter headlines.
2. Elaborate accessories.
3. Remove notes from miro board.
4. Add (T) indicators to structure and headlines.

**Screenshots**
-

**System Information (please complete the following information):**
-

**Additional context**
-

PR #274, opened 2024-12-20
Human structure needs some fixes

ipg-sig:
The human structure table features very specific node names (e.g. Accessories_Wristband).
The term accessories should be elaborated. It should be made clear that it covers former Grp_Equipment_Dynamic (If equipment is very specific for a human asset, rather small, not self-contained or animated, it can be part of the human asset.) as individual meshes.
Notes from miro board should be removed.
As bones create local coordinate frames (transforms), those should be indicated in the structure and headlines.

## Describe your changes
- Adjusted structure and example (reparented legs to hip instead of root)
- Adjusted general description to elaborate accessories
- Adjusted "author" and "co-author" to "authors"
- Removed notes from miro boards,
- Added (T) to structure and headlines

https://github.com/asam-ev/OpenMATERIAL/issues/263

## Mention a member
@ClemensLinnhoff @LudwigFriedmann 


content/geometry/object-human/human-structure.adoc
```diff
@@ -8,36 +8,36 @@ legend
...
-            |_ [[../geometry/object-human/human-index.html#_full_toes_left Full_Toes_Left]]
-      |_ [[../geometry/object-human/human-index.html#_upper_leg_right Upper_Leg_Right]]
-        |_ [[../geometry/object-human/human-index.html#_lower_leg_right Lower_Leg_Right]]
-          |_ [[../geometry/object-human/human-index.html#_foot_right Foot_Right]]
-            |_ [[../geometry/object-human/human-index.html#_full_toes_right Full_Toes_Right]]
-    |_ [[../geometry/object-human/human-index.html#_accessories Accessories_Wristband]]
-    |_ [[../geometry/object-human/human-index.html#_body Body_Human]]
-    |_ [[../geometry/object-human/human-index.html#_clothing Clothing_Vest]]
-    |_ [[../geometry/object-human/human-index.html#_hair Hair_Short]]
+      |_ [[../geometry/object-human/human-index.html#_hip Hip]] (T)
```
ClemensLinnhoff, line 11: You need to add _t to all links where you added the (T) in the headline, e.g. ../geometry/object-human/human-index.html#_hip_t. Otherwise the link won't work.

Adding the (T) outside of the link display name is a very good idea. We should do that in vehicle structure and environment structure as well. Would you mind doing that in this PR directly? I don't think we need another issue for this.

content/geometry/object-human/human-structure.adoc
```diff
@@ -8,36 +8,36 @@ legend
 |_ [[../geometry/object-human/human-index.html#_grp_root Grp_Root]]
   |_ [[../geometry/object-human/human-index.html#_armature Armature]]
```
ClemensLinnhoff, line 9: Here in the structure it is just called Armature, but the section headline is Armature_<Name>. The name however is not specified there. So should we just call it Armature without any suffix?

content/geometry/object-human/human-structure.adoc
```diff
@@ -8,36 +8,36 @@ legend
...
+            |_ [[../geometry/object-human/human-index.html#_foot_left Foot_Left]] (T)
+              |_ [[../geometry/object-human/human-index.html#_full_toes_left Full_Toes_Left]] (T)
+        |_ [[../geometry/object-human/human-index.html#_upper_leg_right Upper_Leg_Right]] (T)
+          |_ [[../geometry/object-human/human-index.html#_lower_leg_right Lower_Leg_Right]] (T)
+            |_ [[../geometry/object-human/human-index.html#_foot_right Foot_Right]] (T)
+              |_ [[../geometry/object-human/human-index.html#_full_toes_right Full_Toes_Right]] (T)
+    |_ [[../geometry/object-human/human-index.html#_accessories Accessories_<Name>]]
+    |_ [[../geometry/object-human/human-index.html#_body Body_<Name>]]
+    |_ [[../geometry/object-human/human-index.html#_clothing Clothing_<Name>]]
+    |_ [[../geometry/object-human/human-index.html#_hair Hair_<Name>]]
```
ClemensLinnhoff, line 40:
```suggestion
    |_ [[../geometry/object-human/human-index.html#_accessories_name Accessories_<Name>]]
    |_ [[../geometry/object-human/human-index.html#_body_name Body_<Name>]]
    |_ [[../geometry/object-human/human-index.html#_clothing_name Clothing_<Name>]]
    |_ [[../geometry/object-human/human-index.html#_hair_name Hair_<Name>]]
```

ClemensLinnhoff (changes requested):

ClemensLinnhoff:
I know we have had this discussion before, but shouldn't we change "Pointing upwards" for the bone y-axis to "Pointing along the bone"? I still think this is more correct, because in the A-pose the bones of the arms for example are pointing down not upwards.

content/geometry/object-human/human-structure.adoc
ipg-sig: I would add and declare it as an optional postfix, but as recommend if multiple human assets are prepared in one 3D scene.

ipg-sig:
@ClemensLinnhoff I changed it for the bones. I you think it´s better and doesn´t interrupt the reading flow, we can keept it like that.

ClemensLinnhoff (approved):

Issue #268, opened 2024-12-19
Extend "author" to "authors" in schemata, documentation and examples


LudwigFriedmann:
**Is your feature request related to a problem? Please describe.**
Currently, the field is called "author". We'll have the need to specify multiple authors

**Describe the solution you'd like**
Rename author to authors in all schemata, documentation and examples.

**Describe alternatives you've considered**
-

**Additional context**
-

PR #284, opened 2025-01-07
Allow multiple authors

LudwigFriedmann:
## Describe your changes

Change 'author' field to 'authors' in schemas and examples .
Remove commas from 'copyright' fields to enable comma-separated lists (like in authors) and extend documentation.

#268 

## Mention a member

@ClemensLinnhoff @AsamDiegoSanchez 


AsamDiegoSanchez (approved):
Reviewed

AsamDiegoSanchez:
should we also change the "copywrite" into plural allowing multiple -> "copywrites"?

LudwigFriedmann:
> > should we also change the "copyrights" into plural allowing multiple -> "copyrights"?
> 
> Good point. What do you think @LudwigFriedmann ?

I can change that. Should I also change 'source' to 'sources'?
The data in a look-up table could also be compiled from several sources...

ClemensLinnhoff:
Yes we could also make it sources. But there I would leave it as a free text string instead of changing it to an array.

LudwigFriedmann:
> Yes we could also make it sources. But there I would leave it as a free text string instead of changing it to an array.

Alright I'll apply the following changes:
* author -> authors (string array)
* copyright -> copyrights (string array)
* source -> sources (string)

Concerning the usage of string arrays, I'd like to point out that this solution is much more sensitive to errors (e.g. misplaced commas). A parser would fail or abort because the author list is not formatted correctly. Do we want to achieve this kind of behavior?

ClemensLinnhoff:
> Concerning the usage of string arrays, I'd like to point out that this solution is much more sensitive to errors (e.g. misplaced commas). A parser would fail or abort because the author list is not formatted correctly. Do we want to achieve this kind of behavior?

I see it the other way around. An error in a defined list of strings is easily found by a JSON schema validator. If it is just a string, the separation might be unclear, when trying to parse it automatically.

LudwigFriedmann:
Hi @ClemensLinnhoff , I implemented everything as discussed. Can you commit a review?

ClemensLinnhoff:
The definition of the schemas was actually not correct. You defined the fields as arrays of arrays of strings instead of just arrays of string. I corrected this myself. I also added that the arrays need at least one item, because the fields are mandatory.

So from my point of view this is ready to be merged.

ClemensLinnhoff (approved):

PR #292, opened 2025-01-13
Add description of coordinate transformation to other standards

### LINKED ISSUE #288: Explain coordinate transformations to other ASAM standards
The information of how to get from the OpenMATERIAL 3D asset origin (center of bounding box projected to the ground) to other ASAM object coordinates is briefly described in the documentation, but somewhat hidden.

Add an explicit and more detailed explanation to the coordinate systems section in the Geometry chapter.

ClemensLinnhoff:
## Describe your changes

Add description of coordinate transformation to other standards.
Furthermore, I noticed that the reflectance table was still missing from the main introduction, so I quickly added it here, too.

Fixes #288 

### DIFF
--- a/content/general-docs/introduction.adoc
+++ b/content/general-docs/introduction.adoc
@@ -97,6 +97,7 @@ The defined look-up tables include:
 * xref:material/material-emp-schema.adoc[Electromagnetic properties]
 * xref:material/material-optical-schema.adoc[Optical properties]
 * xref:material/material-brdf-schema.adoc[Bidirectional reflectance distribution functions (BRDFs)]
+* xref:material/material-reflectance-schema.adoc[Reflectance table]
 
 Examples for _material property_ files and look-up table files are provided in the https://github.com/asam-ev/OpenMATERIAL/tree/main/examples/[examples folder] of the repository.
 
--- a/content/geometry/general.adoc
+++ b/content/geometry/general.adoc
@@ -91,6 +91,8 @@ Top-most local coordinate frames of _3D assets_ (so-called reference coordinate 
 By static transformations, these coordinate frames may be synchronized with reference coordinate frames in ASAM OpenSCENARIO and ASAM OSI.
 The following table provides and overview of the translations between the ASAM OpenMATERIAL 3D asset reference coordinates to reference coordinates of ASAM OpenSCENARIO XML and ASAM OSI.
 
+This is a test: <<tab-local-coordinates>>
+
 [#tab-local-coordinates]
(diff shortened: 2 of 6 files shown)


ClemensLinnhoff:
@KimuraDIVP would these explanations answer your questions about the coordinate systems of other ASAM standards? Or is something missing?

ktrmpc:
> @KimuraDIVP would these explanations answer your questions about the coordinate systems of other ASAM standards? Or is something missing?

There is no shortage of information.
I think it would be very helpful if section 7.5.1.15.3 contained an explanation on how to convert the origin position to other ASAM standards (OpenSCENARIO, OSI, etc.).

ClemensLinnhoff:
> I think it would be very helpful if section 7.5.1.15.3 contained an explanation on how to convert the origin position to other ASAM standards (OpenSCENARIO, OSI, etc.).

I would not put explanations about coordinate transformation there, since this is just the json schema for the asset file. Furthermore, the axles are not needed for all conversions, e.g. to the OSI object coordinate frame. That's why I put the explanations to the coordinate systems chapter 7.2.2

FKlopfer (approved):

AsamDiegoSanchez (approved):
reviewed.

ktrmpc:
> > I think it would be very helpful if section 7.5.1.15.3 contained an explanation on how to convert the origin position to other ASAM standards (OpenSCENARIO, OSI, etc.).
> 
> I would not put explanations about coordinate transformation there, since this is just the json schema for the asset file. Furthermore, the axles are not needed for all conversions, e.g. to the OSI object coordinate frame. That's why I put the explanations to the coordinate systems chapter 7.2.2

Thank you for your reply.
I understood the explanation in section 7.2.2. The contents of section 7.2.2 are very good, but to make it even better, I think that if the OpenMATERIAL3D definition is compatible with the ASAM standard, it will be widely used by users. Therefore, I thought it would be good to show that this schema can be used to easily convert the origin and coordinate system when OpenMATERIAL3D works with other ASAM standards. After reading this chapter, especially up to section 7.5.1.15.3, I thought that users might be worried about whether the origin is compatible with other standards.

ClemensLinnhoff:
Okay. I added a description in the axle description of the asset schema and I additionally added a link  to the vehicle class general section.

ktrmpc:
> Okay. I added a sentence in the axle description of the asset schema and I additionally added a link in the vehicle class general section.

It has become easier to understand how to link to other ASAM standards. Thank you!

PR #294, opened 2025-01-14
More detailed mesh count description

### LINKED ISSUE #293: Mesh count mismatch in glTF
**Describe the bug**
The [meshCount](https://asam-ev.github.io/OpenMATERIAL/asamopenmaterial/latest/specification/geometry/asset-schema.html#_meshcount) a metadata property of an asset. If a mesh has multiple materials, during glTF export, it is split into individual meshes with one material per mesh. So the mesh count before and after the export to glTF is different.

**Expected behavior**
We need to specify, which number of meshes shall be used as the mesh count:
- How it was modeled
- How it ended up in the 3D model file

**Additional context**
I only know this problem exists for glTF files. But it might also exist for FBX and USD.

ClemensLinnhoff:
## Describe your changes

Add more details to the mesh count description due to potentially changing number of meshes during export of the model.

Fixes #293 

### DIFF
--- a/schemas/asset_schema.json
+++ b/schemas/asset_schema.json
@@ -234,7 +234,7 @@
                 },
                 "meshCount": {
                     "type": "integer",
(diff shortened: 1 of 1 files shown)


LudwigFriedmann (approved):
Looks good to me!

PR #326, opened 2025-01-28
Removed individual coordinate systems, added bounding boxes for objects

### LINKED ISSUE #313: Environment structure: Object grouping and coordinate systems unclear
**Describe the bug**
Texts and illustrations in documentation chapter 7.3.4 Environment Structure, more precisely
- Grp_Terrain
- Grp_Environment_Objects
- Grp_Buildings
- Grp_Vegetation
- Grp_Road_Network
- Grp_Sidewalks
- Grp_Road_Marks
- Grp_Road_Objects
- Grp_Signals

are potentially misleading.

Text and iluustrations read as the corresponding group nodes describe single objects.
The illustrations even show individual coordinate systems.
However, this is not the case. The group nodes share their origin with Grp_Root, i.e. the reference coordinate system.
We have to decide here whether we are talking about individual objects or overarching groups here.
In the former case, we may have to define coordinate systems. 

**To Reproduce**
1. See Documentation 7.3.4 Environment Structure

**Expected behavior**
We have to decide here whether we are talking about individual objects or overarching groups.
In the former case, we would have to define coordinate systems.
In the second case, we have to adapt the texts and illustrations and state that the objects share the coordinate frame of Grp_Root.

**Screenshots**
-

**System Information (please complete the following information):**
 -

**Additional context**
-

ACMackenthun:
## Describe your changes

Remove individual coordinate systems from environment object images and add bounding boxes for objects.

**Some questions to ask**:

- Coordinate systems in images where misleading and need to be removed to avoid misunderstandings.

Fixes #313 


content/_images/Grp_Road_Network.svg
ACMackenthun, line 1: @ClemensLinnhoff: I added the label to indicate that it refers to the complete road network, including the drivable area and the sidewalks. However, I wonder if that is sufficient. Should I add other traffic objects like signals as well?

content/_images/Grp_Road_Objects.svg
ACMackenthun, line 1: @ClemensLinnhoff: Do you agree with the representation of the bounding boxes? I had to change the color of the poles to visualize the pole bounding boxes, but they are still hard to see. I could also add labels like in the Grp_Signals image (https://asam-ev.github.io/OpenMATERIAL/asamopenmaterial/latest/specification/_images/Grp_Signals.svg), or extend the spec and state that the bounding boxes in the image represent the road objects

content/_images/Grp_Terrain.svg
ACMackenthun, line 1: @ClemensLinnhoff: Without the coordinate system the image looses its information value. Do we need the image at all or should we come up with an image with a different perspective?

If we draw a new image, e.g. with a hill next to the road in https://asam-ev.github.io/OpenMATERIAL/asamopenmaterial/latest/specification/_images/Grp_Buildings.svg, it takes about 1-2 hours, because we should probably add a 3D object with shaded triangles and try out different grayscales.

content/_images/Grp_Road_Objects.svg
ACMackenthun: BTW: Do we also need to remove the coordinate systems from Grp_Sign and Grp_Traffic_Light?

content/_images/Grp_Buildings.svg
ClemensLinnhoff, line 1: Could you add a bounding box to the building? I mean in the lower part bounding box and bounding are the same, but at the roof the bounding box would go straight up. (Same in the Grp_Environment_Objects)

content/_images/Grp_Road_Network.svg
ClemensLinnhoff: Yes, maybe it's a good idea to add just one road object and maybe one signal to show that they are also part of the road network.

content/_images/Grp_Road_Objects.svg
ClemensLinnhoff: In my opinion the bounding boxes for the poles are fine.

The coordinate frames for signs and traffic lights should remain, as they refer to individual objects with their own transform.

content/_images/Grp_Terrain.svg
ClemensLinnhoff: Hm good point. Maybe it would make more sense to have a road which is grayed out and a green area left and right from the road. And then you could add the label and arrows to the terrain, similar to e.g. the sidewalk.

ClemensLinnhoff (changes requested):

content/_images/Grp_Buildings.svg
ACMackenthun: Done, also changed the line width of the bounding box to enhance visibility.

Even though I liked the transparency of the original building, I decided to overlay the bounding box lines with the building because the building looked crooked with the higher bounding box, see 

image

ACMackenthun: TODO: remove bounding box line from left edge of image or move building slightly to the back to visualize complete box

content/_images/Grp_Road_Network.svg
ACMackenthun: done

content/_images/Grp_Terrain.svg
ClemensLinnhoff: But now everything is gray and it is not clear what belongs to the Terrain.

ACMackenthun: Right, what do you think about this new version with the green hills and the labels?

ClemensLinnhoff: That looks good, thanks.

ClemensLinnhoff (approved):

PR #329, opened 2025-01-28
Checker Rules

### LINKED ISSUE #280: Do we need checker rules?
#### Describe the problem/question

Do we need checker rules that can be implemented later to be used i.e in the Quality Checker Framework?
(see in OpenDRIVE (https://publications.pages.asam.net/standards/ASAM_OpenDRIVE/ASAM_OpenDRIVE_Specification/latest/specification/16_annexes/map_rules.html) and OpenSCENARIO XML)

#### Describe your research

The ASAM Quality Checker Framework (https://github.com/asam-ev/qc-framework) operates through a series of specialized "checkers," organized into checker libraries. These checkers evaluate different aspects of your files, ensuring they meet the required quality standards.

### LINKED ISSUE #343: Annexes can be normative or informative: annex heading must contain the indication "(normative)"?

Do we want the annexes to be already "normative"?

If yes, we need to specifically mention it like in OpenDRIVE (https://publications.pages.asam.net/standards/ASAM_OpenDRIVE/ASAM_OpenDRIVE_Specification/latest/specification/index.html)

Image

If yes, add "(normative)" text to the annexes

ClemensLinnhoff:
## Describe your changes

Add checker rules to annex

Fixes #280, fixes #343


ClemensLinnhoff:
@DavidJRitter904 I set up the annex page with a description of the concept and the rules we defined in the meeting today.
I added a note at the end, that rules for all physical properties should be added. So feel free to add the rules right here on this branch.

DavidJRitter904:
I have added rules for 12  physical properties. 

1. wavelength in m
2. real part index of refraction (refractive index n)
3. imaginary part index of refraction (extinction coefficient k)
4. humidity in %
5. real + imaginary part permeability
6. real + imaginary part permittivity
7. incident + exit elevation angle in rad
8. azimuth angle in rad
9. BRDF in 1/sr
10. polarized plane angle in rad
11. amplitude linearly polarized plane 
12. phase linearly polarized plane in rad

For most of them I have considered the theoretical possible ranges and narrowed them down to a range which I thought is in accordance to our use cases. 
If I have missed a property, please comment here and I will include it.

Feedback is appreciated!

content/appendix/checker-rules.adoc
```diff
@@ -2,17 +2,201 @@
...
+
+UID:: asam.net:xom-mat:1.0.0:xompt.tables_sorted_correctly
+
+Description:: Arrays in look-up tables shall be sorted based on the columns starting with the first.
+
+==== wavelength_valid_range
+
+UID:: asam.net:xom-mat:1.0.0:xompt.wavelength_valid_range
+
+Description:: The wavelength shall be within the range of 1e-09 m to 17,16e-03 m.
```
ClemensLinnhoff, line 118:
```suggestion
Description:: The wavelength shall be within the range of 1e-09 m to 17.16e-03 m.
```

Why is 17.16e-03 m the maximum?

content/appendix/checker-rules.adoc
```diff
@@ -2,17 +2,201 @@
...
+
+UID:: asam.net:xom-mat:1.0.0:xompt.temperature_valid_range
+
+Description:: A temperature property shall not be below 0 K.
+
+==== real_part_IOR_valid_range
+
+UID:: asam.net:xom-mat:1.0.0:xompt.real_part_IOR_valid_range
+
+Description:: The real part of index of refraction (refractive index n) shall be below 0.
```
ClemensLinnhoff, line 130: Why below 0? Shouldn't it be above? Glass for example has an IOR of about 1.5.

content/appendix/checker-rules.adoc
```diff
@@ -2,17 +2,201 @@
...
+
+UID:: asam.net:xom-mat:1.0.0:xompt.imaginary_part_IOR_valid_range
+
+Description:: The imaginary part of index (extinction coefficient k) of refraction shall not be below 0.
+
+==== humidity
+
+UID:: asam.net:xom-mat:1.0.0:xompt.humidity
+
+Description:: The humidity shall be within the range of 0 % to 100 %.
```
ClemensLinnhoff, line 142: I would write between 0 and 1 to not cause confusion with the %.

content/appendix/checker-rules.adoc
DavidJRitter904: For the uppper limit of the wavelength, I have taken 20 kHz which is range where ultrasound begins. Since we have also considered it in our discussions, I did not want to exclude it.
20 kHz result to 17.16e-03 m for wavelength.

content/appendix/checker-rules.adoc
DavidJRitter904: This would cause confusion as we defined it with percentage within the documentation and the example given also has the value "50" which is why I have chosen 0 - 100.

content/appendix/checker-rules.adoc
DavidJRitter904: There is indeed a "not" missing. Thank you!

content/appendix/checker-rules.adoc
ClemensLinnhoff: That makes sense. I added a sentence to explain this.

content/appendix/checker-rules.adoc
ClemensLinnhoff: I added the "not"

content/appendix/checker-rules.adoc
ClemensLinnhoff: You are right. But then the range should be 0 - 100. So I would remove the %, as e.g. 50 % == 0.5.

ClemensLinnhoff:
@SandroReith could you also review this?

ClemensLinnhoff:
The properties in XOMP files are not part of this, yet. If we want to add more rules for XOMP, I would propose to do it in a separate Issue/Branch.

SandroReith (approved):

SandroReith:
> I have added rules for 12 physical properties.
> 
> 1. wavelength in m
> 2. real part index of refraction (refractive index n)
> 3. imaginary part index of refraction (extinction coefficient k)
> 4. humidity in %
> 5. real + imaginary part permeability
> 6. real + imaginary part permittivity
> 7. incident + exit elevation angle in rad
> 8. azimuth angle in rad
> 9. BRDF in 1/sr
> 10. polarized plane angle in rad
> 11. amplitude linearly polarized plane
> 12. phase linearly polarized plane in rad
> 
> For most of them I have considered the theoretical possible ranges and narrowed them down to a range which I thought is in accordance to our use cases. If I have missed a property, please comment here and I will include it.
> 
> Feedback is appreciated!

Hello, I find the overview excellent. A few general questions, about point one. Shouldn't we check against Si units right
away. To point 4, do we still need the reference value here? So how much is 30% humidity - that this value was determined for the material at 30 g/m^3? [DIN EN ISO 1110:2019-09, DIN EN ISO 291:2008-08]
Do we agree on a coordinate system and can check against it?
We talked about it last time with the information, but can you also check e.g. if the description has gaps? When it comes to frequency, I wouldn't think of why it might make sense to have non-linear or non-equidistant step widths in a band. 

Thanks for the effort @DavidJRitter904. I like it very much.

ClemensLinnhoff:
> We talked about it last time with the information, but can you also check e.g. if the description has gaps? When it comes to frequency, I wouldn't think of why it might make sense to have non-linear or non-equidistant step widths in a band.

In my opinion this makes sense for lidar, e.g. 905 nm, 1310 nm, 1550 nm.

SandroReith:
> In my opinion this makes sense for lidar, e.g. 905 nm, 1310 nm, 1550 nm.

Even when we think of FMCW?

ClemensLinnhoff:
> > In my opinion this makes sense for lidar, e.g. 905 nm, 1310 nm, 1550 nm.
> 
> Even when we think of FMCW?

No, mainly for ToF. But for FMCW you would still have distinct bands, something like this: 900, 905, 910, 1300, 1305, 1310, 1315, 1320

DavidJRitter904:
1. Where applicable I have added the units, such as "rad" or (1/sr) for the rules definition as well as for humidity (%). While implementing the checker itself, the units can be neglected. My idea is to make it clear in the set of rules what units we are referencing our valid range to. 
Either we want to include units in all rules or neglect them for all paramater. What do you think?

2. For humidity we take the relative humidity values here, not the absolute ones measuring the mass per volume as @SandroReith  pointed out above referencing to the ISO standards.
Otherwise the valid range would not have an upper limit.

3. humidity misses the "_valid_range" part in the name

FKlopfer (approved):

ClemensLinnhoff:
I will fix humidity unit.

@DavidJRitter904 will add value ranges for properties in the XOMP file schema.

PR #361, opened 2025-02-05
Change chapter order in vehicle chapter

FKlopfer:
## Describe your changes

Change chapter order in vehicle chapter to "mounting - joint - view"

[Add a reference to a related issue.](https://github.com/asam-ev/OpenMATERIAL-3D/issues/358)

### DIFF
--- a/content/geometry/object-vehicle/vehicle-structure.adoc
+++ b/content/geometry/object-vehicle/vehicle-structure.adoc
@@ -1240,42 +1240,6 @@ image::Grp_Light_Warning.svg[,600]
 | Perpendicular to the x-axis, pointing vertically upwards
 |===
 
-
-== Grp_Mirror_Blindspot_Joint_<blindspot_mirror_joint_idx> (T)
-
-This group contains all parts of the movable structure that holds the blindspot mirror.
-The blindspot mirror view automatically adjusts when the angle of the blindspot joint changes.
-
-It is a child node of the corresponding mirror mounting group.
-
-`<blindspot_mirror_joint_idx>` denotes the index of blindspot mirror joints. The index entries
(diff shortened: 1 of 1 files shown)


ClemensLinnhoff (approved):

content/geometry/object-vehicle/vehicle-structure.adoc
```diff
@@ -1386,30 +1385,34 @@ image::Grp_Mirror_Side_Joint_Left.svg[,600]
 |===
 
 
-== Grp_Mirror_Side_Joint_Right_<side_mirror_joint_right_idx> (T)
+== Grp_Mirror_Side_Mounting_Right_<side_mirror_mounting_right_idx> (T)
```
LudwigFriedmann, line 1388: For this sequence, I'd really follow the structure as in the structure overview:
1. Side mirror mounting left -> Side mirror joint left -> Side mirror view left
2. Side mirror mounting right -> Side mirror joint right -> Side mirror view right

LudwigFriedmann (changes requested):
Looks good, only the sequence of the side mirror nodes is not consistent with the structure overview

LudwigFriedmann:
I just realized Grp_Interior_Static chapter is also not in sequence with the structure overview and the "Optional_Grp_Vehicle_Part" in the structure overview should be changed to "Grp_Vehicle_Part". At least this is how we handle optional nodes in all other structures

content/geometry/object-vehicle/vehicle-structure.adoc
FKlopfer: Updated chapter sequence and removed "optional" from structure overview

LudwigFriedmann (approved):
Looks good to me!

PR #389, opened 2025-02-19
Move title to the right in the header

FKlopfer:
## Describe your changes

- Changed title in the Header from "OpenMATERIAL 3D" to "ASAM OpenMATERIAL 3D"
- Move title to the right 
- Use "supplemental_ui" (https://docs.antora.org/antora/latest/playbook/ui-supplemental-files/)

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -2,7 +2,7 @@
 # Source: https://code.asam.net/common/templates/antora-base-repo-example/-/blob/standard_generator/site.yml
 # tag::site[]
 site:
-  title: OpenMATERIAL® 3D
+  title: ASAM OpenMATERIAL® 3D
   url: https://asam-ev.github.io/OpenMATERIAL-3D/
   start_page: asamopenmaterial:specification:index.adoc 
 # end::site[]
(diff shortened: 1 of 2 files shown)


FKlopfer:
**Currently:** 
image

**New:**
image

AsamDiegoSanchez (approved):
Reviewed.
My only concern right now is if this will affect the Pastel link where the public review is now on-going...Pastel is taking the HTML version of it: https://asam-ev.github.io/OpenMATERIAL-3D/asamopenmaterial/latest/specification/index.html

ClemensLinnhoff:
> Reviewed. My only concern right now is if this will affect the Pastel link where the public review is now on-going...Pastel is taking the HTML version of it: https://asam-ev.github.io/OpenMATERIAL-3D/asamopenmaterial/latest/specification/index.html

I don't think it will affect pastel because the url does not change, right?

ClemensLinnhoff:
This needs to be done for all the ASAM standards, right? So why is the supplemental_ui hardcoded here. Shouldn't it be a submodules from a central ASAM repository or part of the antora pipeline?

AsamDiegoSanchez:
> > Reviewed. My only concern right now is if this will affect the Pastel link where the public review is now on-going...Pastel is taking the HTML version of it: https://asam-ev.github.io/OpenMATERIAL-3D/asamopenmaterial/latest/specification/index.html
> 
> I don't think it will affect pastel because the url does not change, right?

Then we can merge it. And anyway this could be just a temporary adjustment (workaround) in OpenMATERIAL 3D till I align internally with all the other standards that will be affected if we finally modify the general Antora UI

AsamDiegoSanchez:
> This needs to be done for all the ASAM standards, right? So why is the supplemental_ui hardcoded here. Shouldn't it be a submodules from a central ASAM repository or part of the antora pipeline?

Yes, but I have to bring up to speed the other standard TMs. This is just a temporary workaround to present the idea.

ClemensLinnhoff:
Okay

Issue #317, opened 2025-01-27
Release process and versioning


ClemensLinnhoff:
#### Describe the problem

At the end of this project we need to do a release 1.0.0.
I would assume we use the GitHub release functionality and set a release tag v1.0.0 to the last commit in the main branch at that time.
Since at that point there will be only one version of the standard, there will also only be one version of the documentation.
However, over time, there will be a current development in the main branch after 1.0.0 and later also future version.

#### Describe your research

In OSI you can select different versions of the standard and also the current development state on the upper right side of the documentation page,  see screenshot:

![Image](https://github.com/user-attachments/assets/eafd3789-e09d-42fe-a79c-f4e667352457)

We seem to have something similar in the OpenMATERIAL 3D documentation on the lower left side. But this currently says 1.0.0, which is not released, yet.

![Image](https://github.com/user-attachments/assets/ccb7371a-45fe-4621-aba7-6f145d9a95f3)

I think the main issue will be the automatically generated ASCIIdoc files from the JSON schemas. Currently, this is done by calling a python script in the build pipeline [here](https://github.com/asam-ev/OpenMATERIAL/blob/main/.github/workflows/antora-build.yml#L48). 

#### Ask your question

How will this work with multiple versions of the standard?

martinfiebig:
Hi @ClemensLinnhoff, with Antora, there are basically two ways to have different versions of the documentation, by using branches or by using tags to store the older versions of the content. Both ways, the documentation will then be generated from the old version(s) of the content. This should also work for the generation of the AsciiDoc files from JSON - I will test this in my forked repository and report back.

martinfiebig:
Hi @ClemensLinnhoff, I've tested the handling of multiple versions in my fork and unfortunately, the generation of the AsciiDoc files from JSON does not work for older versions, only for the current one. So for the older versions, the files would need to be generated elsewhere and then be stored in the respective branches of the older versions.

ClemensLinnhoff:
What would you recommend what would be the best way to do this?
How does it for example work with the OSI Doxygen documentation, which is also generated from code? Is this somethow done in the antora image at ASAM?

martinfiebig:
Hi Clemens, we will look into the workflow for OSI. The antora build in OSI is however much more complicated than in OpenMATERIAL, since OSI uses multiple repositories to store and generate the documentation. So I'm not quite sure, if we can take over the mechanisms used in OSI.
We will test that and come back with possible solutions.

[...]

PR #413, opened 2025-03-11
317 release process and versioning

martinfiebig:
## Describe your changes

Added github workflow to build and commit the asciidoc files automatically generated from schema files.
Included tags in the build workflow.

#317 

### DIFF
--- a/.antora/antora.yml
+++ b/.antora/antora.yml
@@ -2,16 +2,7 @@
 name: asamopenmaterial
 # CENTRAL DOCUMENT ONLY!!! Display name of the component. Define only once per component!
 title: ASAM OpenMATERIAL 3D
-version:
-  # Update version number accordingly. This is the version for the main branch
-  main: '1.0.0'
-  master: '1.0.0'
-  v(*)-antora: V$1
-  x-antora-v(*): V$1
-  v(*): V$1
-  V(*): V$1
-  # Update version number accordingly. This is the version for any other activated branch (branch must be added in the generator site.yml!). Schema = 0.0.x-<branch-name> (only works for branches starting with "feature/")
-  (*): 'V1-0-0-$1'
+version: 'draft'
 # CENTRAL DOCUMENT ONLY!!! First page of the component. Define only once per component!
 start_page: specification:index.adoc
 
--- a/.antora/antora.yml
+++ b/.antora/antora.yml
@@ -4,7 +4,7 @@ name: asamopenmaterial
 title: ASAM OpenMATERIAL 3D
 version: 
   v(?<v>+({0..99}).+({0..99}).+({0..99})): $<v>
-  deploy: 'current'
+  main: 'current'
 # CENTRAL DOCUMENT ONLY!!! First page of the component. Define only once per component!
 start_page: specification:index.adoc
 
--- a/.antora/antora.yml
+++ b/.antora/antora.yml
@@ -3,7 +3,7 @@ name: asamopenmaterial
 # CENTRAL DOCUMENT ONLY!!! Display name of the component. Define only once per component!
 title: ASAM OpenMATERIAL 3D
 version: 
-  release-v(?<v>+({0..99}).+({0..99}).+({0..99})): $<v>
+  v(?<v>+({0..99}).+({0..99}).+({0..99})): $<v>
   main: 'current'
 # CENTRAL DOCUMENT ONLY!!! First page of the component. Define only once per component!
 start_page: specification:index.adoc
--- a/.antora/antora.yml
+++ b/.antora/antora.yml
@@ -5,7 +5,7 @@ title: ASAM OpenMATERIAL 3D
 version: 
   v(?<v>+({0..99}).+({0..99}).+({0..99})): $<v>
   main: 'current'
-  (*): 'pull-request-$1'
+  (*): 'pull-request'
 # CENTRAL DOCUMENT ONLY!!! First page of the component. Define only once per component!
 start_page: specification:index.adoc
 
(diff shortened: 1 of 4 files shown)


.antora/antora.yml
```diff
@@ -2,16 +2,7 @@
...
-  # Update version number accordingly. This is the version for the main branch
-  main: '1.0.0'
-  master: '1.0.0'
-  v(*)-antora: V$1
-  x-antora-v(*): V$1
-  v(*): V$1
-  V(*): V$1
-  # Update version number accordingly. This is the version for any other activated branch (branch must be added in the generator site.yml!). Schema = 0.0.x-<branch-name> (only works for branches starting with "feature/")
-  (*): 'V1-0-0-$1'
+version: 'draft'
```
ClemensLinnhoff, line 5: I would rename it to 'current' to be consistent with ASAM OSI.

antora-playbook.yml
```diff
@@ -18,6 +18,9 @@ content:
     tags: [v*,V*]
   - url: .
     start_path: .antora
+    tags: [v*.0.0]
+    #tags: [v*]  -> change to this to also build and include minor versions
```
ClemensLinnhoff, line 22: I would include all release including minor and patch releases.

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+         submodules: recursive
+
+    - name: build files and push changes to remote
+      working-directory: repo/scripts
+      run: |
+        git checkout deploy
+        python3 generate_schema_docs.py
+        git config --global user.name "Commit from Workflow"
+        git config --global user.email "test@test.com"
+        git add ../content/*schema.adoc -f
```
ClemensLinnhoff, line 35: Are you sure this will work? Because the generated schema adoc files are in subfolders within the content folder. Is this covered by the *?

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+        git commit -m "Automatically commit generated AsciiDoc files"
+        git push
```
ClemensLinnhoff, line 37: @AsamDiegoSanchez could you create the deploy branch? It should be protected, so users cannot push to it directly.
Does this pipeline have the rights to push to the deploy branch directly?

.antora/antora.yml
```diff
@@ -2,16 +2,7 @@
...
-version:
-  # Update version number accordingly. This is the version for the main branch
-  main: '1.0.0'
-  master: '1.0.0'
-  v(*)-antora: V$1
-  x-antora-v(*): V$1
-  v(*): V$1
-  V(*): V$1
-  # Update version number accordingly. This is the version for any other activated branch (branch must be added in the generator site.yml!). Schema = 0.0.x-<branch-name> (only works for branches starting with "feature/")
-  (*): 'V1-0-0-$1'
```
ClemensLinnhoff, line 14: Is it possible to add an expression that every release tag from the deploy branch is automatically used as well?

ClemensLinnhoff (changes requested):

.github/workflows/release.yml
AsamDiegoSanchez: "deploy" branch has been created, I will add the protection rules after the meeting

.github/workflows/release.yml
martinfiebig: Yes, I've tried this in my forked repository with the same structure and it worked.

.antora/antora.yml
martinfiebig: renamed to 'current'

.antora/antora.yml
ClemensLinnhoff: Okay but if we only add tags to the deploy branch it would work and the antora.yml does not have to be changed for new releases?

.github/workflows/release.yml
AsamDiegoSanchez: I´ve added the following rules:

image

image

image

should this be enough?

ClemensLinnhoff: I think this will not work. The CI pipeline as it is set up here directly pushes to the deploy branch. It is not done via a pull request.

.antora/antora.yml
martinfiebig: Alright, I just removed the 'release-' prefix :)

ClemensLinnhoff:
The only remaining issue is the protection of the deploy branch. Once that is solved, we can merge.

AsamDiegoSanchez:
> The only remaining issue is the protection of the deploy branch. Once that is solved, we can merge.

I´ve added protection rules to the deploy branch (see my previous comment)

ClemensLinnhoff:
> > The only remaining issue is the protection of the deploy branch. Once that is solved, we can merge.
> 
> I´ve added protection rules to the deploy branch (see my previous comment)

Yes but this will not work, see my comment ;)

AsamDiegoSanchez:
> > > The only remaining issue is the protection of the deploy branch. Once that is solved, we can merge.
> > 
> > 
> > I´ve added protection rules to the deploy branch (see my previous comment)
> 
> Yes but this will not work, see my comment ;)

ah, sorry I oversaw it. From the GitHub settings we could add the following as well:
image

or create additional checks with GitHub actions?:
image

.github/workflows/release.yml
ClemensLinnhoff: I tried something in this test repo: https://github.com/ClemensLinnhoff/deploy_branch_test

We can define a deploy key and use that in the CI Pipeline. Deploy keys can be added to the bypass list in the branch protection rules.
So this generally would work. But I see one security risk: Anyone could use this key in a new CI pipeline to push anything to the deploy branch. So indirectly, every user could just add something to the official documentation. This is not good.

ClemensLinnhoff:
The security risk is actually somewhat mitigated, because GitHub by default restricts secrets to be accessed from forks. So only people from the contributor list may maliciously manipulate the CI pipeline in which case they could be removed from the contributors list.

For me this would be an acceptable risk. That do you think @AsamDiegoSanchez ?

ClemensLinnhoff (approved):

PR #432, opened 2025-03-13
Add deliverables part to navigation bar

FKlopfer:
## Describe your changes

Add deliverables part to navigation bar

[Add a reference to a related issue.](https://github.com/asam-ev/OpenMATERIAL-3D/issues/346)


FKlopfer:
I added the "Deliverable" part to the navigation bar. Currently, there is no package linked to "1.0.0-rc" as this has not yet been created, right? When this is done, the url only needs to be added in the "href" attribute.

ClemensLinnhoff:
But this is different from how it is in OpenDRIVE. It does not change depending on the selected version, right? This is what I asked in the corresponding issue.

AsamDiegoSanchez (approved):
Reviewed.

ClemensLinnhoff (approved):

PR #434, opened 2025-03-17
add scheme to include rc tags

martinfiebig:
## Describe your changes

Adapted the naming scheme to allow the build of release tags (i.e. "v1.0.0-rc1")

### DIFF
--- a/.antora/antora.yml
+++ b/.antora/antora.yml
@@ -5,9 +5,8 @@ title: ASAM OpenMATERIAL 3D
 version: 
   v(?<v>+({0..99}).+({0..99}).+({0..99})): $<v>
   v(?<version>+({0..99}).+({0..99}).+({0..99}))-*: $<version>-rc
-  #v*.*.*-*: release-candidate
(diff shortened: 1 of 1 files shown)


martinfiebig:
Hi @ClemensLinnhoff, I've adapted the naming scheme of the tags to include the '-rc' suffixes as well. I'm not sure, why the local build on this PR fails, but it worked when triggering the build on the main branch:
image
I will delete my test-tag now - and once this PR is merged, we can create the official release tag again.

ClemensLinnhoff:
Can you please figure out, why the local build fails? It needs to work.

martinfiebig:
Hi @ClemensLinnhoff, looking at the logs, there were duplicate nav-files - which indicate that the version names of 2 or more version were the same. I've now put the suffix of the tag in a separate variable and added it to the version name to prevent this from happening in the future. I've also deleted my test tags, so that we can set the release tag as soon as this PR gets merged.

ClemensLinnhoff (approved):

PR #437, opened 2025-03-20
Fix angle definition in BRDFs and reflCoeff

### LINKED ISSUE #436: Possibly wrong definition of BRDF azimuth angle
**Describe the bug**
We have defined the azimuth angle as follows:

> Column 4: Exit azimuth angle in rad. It shall be within a range of 0 to 2π. For BRDFs with incident and exit vectors on the same plane as the normal, exit azimuth angle is 0.

My issue is with the second half of that definition. If incident and exit angles are on the same plane, the azimuth should be 0 on the one half and π on the other, right? So a specular exit angle for example will be with azimuth = π, because it is on the other side of the normal vector.

@KimuraDIVP  @lyndyRott @DavidJRitter904 could you quickly comment on this? We might be able to still fix this before the release.

ClemensLinnhoff:
## Describe your changes

Fix angle definition in BRDFs and reflCoeff

image

Fixes #436 


schemas/material_reflCoeff_schema.json
```diff
@@ -106,25 +106,25 @@
...
+                                "minimum": 0,
                                 "maximum": 1.570796
                             },
                             {
                                 "type": [
                                     "number",
                                     "null"
                                 ],
-                                "description": "Exit azimuth angle in rad. It shall be within a range of \\(0\\) to \\(2 \\pi\\). For BRDFs with incident and exit vectors on the same plane as the normal, exit azimuth angle is 0.",
+                                "description": "Exit azimuth angle in rad. It shall be within a range of \\(0\\) to \\(2 \\pi\\). The value is relative to the incident azimuth angle.",
```
DavidJRitter904, line 127: Add at the end: `(=0°)`
OR
 add: `The incident azimuth angle is set to 0° as the incident ray is used as reference.`

schemas/material_brdf_schema.json
```diff
@@ -108,25 +108,25 @@
...
+                                "minimum": 0,
                                 "maximum": 1.570796
                             },
                             {
                                 "type": [
                                     "number",
                                     "null"
                                 ],
-                                "description": "Exit azimuth angle in rad. It shall be within a range of \\(0\\) to \\(2 \\pi\\). For BRDFs with incident and exit vectors on the same plane as the normal, exit azimuth angle is 0.",
+                                "description": "Exit azimuth angle in rad. It shall be within a range of \\(0\\) to \\(2 \\pi\\). The value is relative to the incident azimuth angle.",
```
DavidJRitter904, line 129: Add at the end: `(=0°)`
OR
 add: `The incident azimuth angle is set to 0° as the incident ray is used as reference.`

lyndyRott (commented):
> The value is relative to the incident azimuth angle.
Should we add  
_, which is defined to be zero in this context._

DavidJRitter904 (changes requested):
I think adding that the incidence ray is azimuth angle = 0° makes it easy to understand that it is the reference we are defining the exit azimuth angle from.

schemas/material_reflCoeff_schema.json
ClemensLinnhoff: I added the sentence.

ClemensLinnhoff:
@KimuraDIVP Does this updated definition also work for you?

DavidJRitter904 (approved):

AsamDiegoSanchez (approved):
Reviewed

KimuraDIVP:
As for Column 4, we stick to "0 to π", not "0 to 2π", to enhance one and only expression. 

Column 4: Exit azimuth angle in rad. It shall be within a range of 0 to π.

lyndyRott (approved):

lyndyRott:
@KimuraDIVP 
Kimura-san, thanks for your feedback!
I think it is no problem if you use BRDFs that only have exit-azimuth in the range of [0, $$\pi$$ ]. But it is better to cover the entire range possible [0, $$\pi$$ ), for pre-existing BRDFs that also evaluate angles up to 2 $$\pi$$. This way we are in agreement with the general definition.

PR #439, opened 2025-03-20
Replace defective osi trace file

### LINKED ISSUE #438: Bug in example trace file
**Describe the bug**
There is something wrong with the example OSI trace file. I am not fully sure what but I have an idea. I think this has been created with an older version of esmini which was not fully standard compliant to OSI. So I should be able to just simulate the scenario again with a newer version of esmini and this should be fixed.

@AsamDiegoSanchez I will have a look at it right away today and hopefully I can fix it and directly create a merge request so we can still have this correct in the release.

ClemensLinnhoff:
## Describe your changes

The trace file had a bug that stationary objects were only present in the first time step. This is now fixed.

Fixes #438 


AsamDiegoSanchez (approved):
Reviewed

PR #463, opened 2026-04-07
462 Remove type requirement for customProperties

ipg-jonas-schwab:
## Describe your changes

Remove the requirement that all customProperties entries must be of type string.


ClemensLinnhoff (approved):
Looks good. I cannot remember, why we set this to string in the schema and it also does not make much sense to me.

KimuraDIVP (approved):

PR #472, opened 2026-05-07
Allow reassignment of mapping keys in Open Material asset file

### LINKED ISSUE #451: Allow the xoma file to assign a material to a single rgb code instead of a texture
If the material of a 3D asset only uses a single ***xomp***-file one has currently two options:
1. map its name (from the 3D data file) in the ***xomm*** file
2. assign a single-color texture in the ***xoma*** file with a rgb code present in the ***xomm*** file

The first requires all of these material names to be explicitly stated in the ***xomm*** file.
This limits the otherwise supported use case of sharing the ***xomm*** file between multiple assets.

The latter only requires a single rgb code per ***xomp*** file.
This would be preferable for ease of reassignment and maintainability.
However, it also has implications concerning performance and resource usage.

A better solution would be to also allow directly assigning an rgb code in the ***xoma*** file. 
```json
    "materialTextureAssignment": [
        [
            "Material_Cube",
            "checker_xom.png"
        ],
        [
            "Material_Sphere",
            "rgb:255;0;0"
        ]
    ]
```
An added benefit would be the ability to generate consistent for false-color images for assets that share a ***xomm*** file.

___
As an alternative with improved backward compatibility and naming:
```json
    "materialTextureAssignment": [
        [
            "Material_Cube",
            "checker_xom.png"
        ],
    ]
    "materialRgbAssignment": [
        [
            "Material_Sphere",
            "rgb:255;0;0"
        ]
    ]
```

ipg-jonas-schwab:


examples/example_mapping.xomm
```diff
@@ -12,7 +12,6 @@
         "creationDate": "20240703T101728Z"
     },
     "materialMapping": [
-        ["Material_Sphere", "example_material.xomp", "white aluminum"],
```
ClemensLinnhoff, line 15: I understand the it makes sense to remove Material_Sphere from the mapping table for the replacement example, but the example_mapping.xomm file is als a generic example of a mapping file. Therefore I think we should keep the Material_Sphere here.
An option would be to add another mesh and material to the example_asset and replace this new material.

ipg-jonas-schwab: So would you propose to add another mesh to the [examples/example_asset.gltf](https://github.com/asam-ev/OpenMATERIAL-3D/blob/main/examples/example_asset.gltf)?
Or should I just add an unused name to the mapping file?
E.g.:
```json
    "materialMapping": [
        ["Material_Cylinder", "example_material.xomp", "white aluminum"],
        ["rgb:255;0;0", "example_material.xomp", "white aluminum"],
        ["rgb:0;11;255", "example_material_2.xomp", "black aluminum"]
    ]
```

ipg-jonas-schwab: @ClemensLinnhoff,
I have currently no means of editing glTF files.
But I added back a mapping via material name (currently not present in the example)

schemas/asset_schema.json
```diff
@@ -341,6 +341,25 @@
                 "maxItems": 2
             }
         },
+        "materialReplacement": {
+            "type": "array",
+            "description": "Optional array containing material reassignments. It links material names contained in the 3D model file to RGB Codes or alternative names used as key for the ASAM OpenMATERIAL 3D material mapping file.",
```
ClemensLinnhoff:
```suggestion
            "description": "Optional array containing material reassignments. It links material names contained in the 3D model file to RGB Codes or alternative names used as key for the ASAM OpenMATERIAL 3D material mapping file. An example of the material replacement mechanism can be found in the examples folder of the ASAM OpenMATERIAL 3D GitHub repository.",
```

LudwigFriedmann (approved):

ClemensLinnhoff (approved):

PR #476, opened 2026-06-01
Add solid angle definition

### LINKED ISSUE #450: Solid angle of detector being required in BRDF metadata
Solid angle of the detector that has been used to measure BRDF data needs to be accessible in the metadata. This should also apply for "assumed" detector solid angles, regarding synthetic BRDF data. In case of primary receiver optics in front of the detector, said solid angle is defined by the optics aperture.

Necessity of stating the receiver/detector solid angle for formatted scattering functions is being directly mentioned in existing standards:
- SEMI ME1392-0116 (2023) - "Guide for angle resolved optical scatter measurements on specular or diffuse surfaces"
- EN ISO 19986:2020 - "Lasers and laser-related equipment - Test method for angle resolved scattering"

ClemensLinnhoff:
## Describe your changes

For rapidly changing BRDFs values over angle, e.g. in very specular or very retro-reflective materials, the solid angle of the receiver during the measurement needs to be known. BRDF values are always defined per solid angle, so they are multiplied by the solid angle of the simulation to get the reflectivity. But if the values are rapidly changing, this can lead to problems with energy conservation, if the solid angle in the simulation is much bigger than the one during measurement.

In this PR, a new optional filed in the metadata of the BRDF look-up table is introduced, so the solid angle of the measurement can be specified.

Fixes #450 

### DIFF
--- a/schemas/material_brdf_schema.json
+++ b/schemas/material_brdf_schema.json
@@ -88,6 +88,30 @@
                         "maximum": 17.16e-03
                     }
                 },
+                "receiverSolidAngles": {
+                    "type": "array",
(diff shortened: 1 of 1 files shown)


LudwigFriedmann (approved):
Looks good to me!

PR #477, opened 2026-06-01
Add addition to the instancing sentence in introduction for variants of assets

### LINKED ISSUE #183: Consideration of OpenSCENARIO DSL Use Cases.
This matter has been temporarily closed in Issue #168, but if we think about i linkage with OpenSCENARIO, I think we’d better allow linking multiple attribute files (asset files in this case) to one 3D shape file. We have launched this issue.
Referring to the scenario structure sample in OpenSCENARIO DSL, the instruction is in the form of "3D shape file + color attributes".

OpenSCENARIO_concrete_scenario_2
OpenSCENARIO_logical_scenario_2
https://publications.pages.asam.net/standards/ASAM_OpenSCENARIO/ASAM_OpenSCENARIO_DSL/latest/conceptual-overview/scenario-abstraction.html#_concrete_scenario

The above means that there is a use case where you want multiple attribute variations for one geometry, and I think that it is different from the current direction of MaterialMapping in OpenMATERIAL, which is to link only one asset files and shape file under the same file name.

It is the same as the previous issue (issue #168), but I would like to re-propose the following two additions so that to make it possible to assign multiple attributes to one shape file.
1.  Make it possible to describe the 3D shape data to be referenced in the asset file. If this is not described, then refer to a 3D shape data file having the same file name.
1.  In the asset file, the assignment information of the material name in the 3D shape data (hereinafter referred to as the 3D material name) and the material name in the material mapping file (hereinafter referred to as the mapping material name) can be described. The 3D material name and the mapping material name without assignment information are treated as the same.

From the point of description, the above two functions are the same as the current specifications unless otherwise stated. 
We believe that supporting such features will improve data efficiency and increase the availability and reusability of test contents.
Material_Mapping_issue_00

ClemensLinnhoff:
## Describe your changes

To facilitate DSL use-case, I added that variants of assets can be defined with a string as a suffix in the file name.

Fixes #183 

### DIFF
--- a/content/00_preface/introduction.adoc
+++ b/content/00_preface/introduction.adoc
@@ -56,7 +56,7 @@ The first interaction with an {THIS_STANDARD}-compliant _3D model_ is through th
 The *asset file* is a JSON file with the file extension .xoma ({THIS_STANDARD} Asset).
 A detailed definition of the https://github.com/asam-ev/OpenMATERIAL-3D/blob/main/schemas/asset_schema.json[asset file JSON schema] is given in xref:07_geometry/asset-schema.adoc[].
 The asset file shall have the same name as a corresponding _3D data file_ in glTF, FBX, or USD format.
-To facilitate instancing, multiple asset files may be linked to a single _3D model_ file by adding an index separated by a dot (.) as a suffix to the asset file name.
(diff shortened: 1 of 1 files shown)


LudwigFriedmann (approved):
Looks good, thank you!

PR #502, opened 2026-08-13
Ensure all information from the schemas are in the specification

### LINKED ISSUE #501: Check if every definition in the schema files is also part of the generated specification.
The schema files have some properties that are not shown in the auto-generated AsciiDoc files in the specification. Close that gap.

_Originally posted by @ClemensLinnhoff in https://github.com/asam-ev/OpenMATERIAL-3D/issues/498#issuecomment-5277631975_

ClemensLinnhoff:
## Describe your changes

- Moved value restrictions from descriptions fields to actual value restrictions and dependecnies in the schemas.
- Rewrote the json2asciidoc script to use all information in the schemas

 Closes #501

### DIFF
--- a/schemas/asset_schema.json
+++ b/schemas/asset_schema.json
@@ -9,18 +9,18 @@
             "properties": {
                 "maxSteering": {
                     "type": "number",
-                    "description": "Maximum steering angle which can be performed by the wheels on this axle. Unit: rad; Range: [0, \\(\\pi\\)], symmetrical.",
+                    "description": "Maximum steering angle which can be performed by the wheels on this axle. Unit: rad. The steering angle is symmetrical in both directions.",
                     "minimum": 0,
                     "maximum": 3.14159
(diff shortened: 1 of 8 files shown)


ipg-jonas-schwab (approved):
:+1: I like the new schema docs.
___
The description of enums has changed slightly
**Enum:** `['vehicle', 'human', 'environment', 'other']` &rarr; **Enum:** `'vehicle', 'human', 'environment', 'other'`
(Both variants are fine to me, I just wondered if it was intentional or not)
___
I also noticed that the mapping schema has stricter constraints concerning the uri.
`^(\./|/)?([a-zA-Z0-9_\-./]+)\.xomp$` &harr; `.*\.xomm$`

I think the optional prefix in the regex is redundant, but the character list does not allow spaces in the path.
Is there a reason, we do not just use `.*\.xomp$` as well?

